PR #50, opened 2021-06-11
Ch808 migrate code in preparation to use clue

jluhrs:
This PR migrates the base Seqexec code to use Java 11 and cats-effect 3, in preparation to use Clue. The main goal was to have it compiling and running. Therefore the code may not be the prettiest. Improvements will be left for a future PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -364,12 +364,12 @@ lazy val observeCommonSettings = Seq(
   // This is important to keep the file generation order correctly
   Universal / parallelExecution := false,
   // Depend on webpack and add the assets created by webpack
-  Compile / packageBin / mappings ++= (webpack in (observe_web_client, Compile, fullOptJS)).value
+  Compile / packageBin / mappings ++= (observe_web_client/Compile/fullOptJS/webpack).value
     .map(f => f.data -> f.data.getName()),
   // Name of the launch script
   executableScriptName := "observe-server",
   // No javadocs
-  mappings in (Compile, packageDoc) := Seq(),
+  Compile/packageDoc/mappings := Seq(),
   // Don't create launchers for Windows
   makeBatScripts := Seq.empty,
   // Specify a different name for the config file
@@ -378,7 +378,7 @@ lazy val observeCommonSettings = Seq(
   bashScriptExtraDefines += """addJava "-javaagent:${app_home}/jmx_prometheus_javaagent-0.3.1.jar=6060:${app_home}/prometheus.yaml"""",
   // Copy logback.xml to let users customize it on site
   Universal / mappings += {
-    val f = (resourceDirectory in (observe_web_server, Compile)).value / "logback.xml"
+    val f = (observe_web_server/Compile/resourceDirectory).value / "logback.xml"
     f -> ("conf/" + f.getName)
   },
   // Launch options
@@ -480,7 +480,7 @@ lazy val app_observe_server_gs_test =
       applicationConfSite := DeploymentSite.GS,
       Universal / mappings := {
         // filter out sjs jar files. otherwise it could generate some conflicts
-        val universalMappings = (mappings in (app_observe_server, Universal)).value
+        val universalMappings = (app_observe_server/Universal/mappings).value
         val filtered          = universalMappings.filter { case (_, name) =>
           !name.contains("_sjs")
         }
@@ -509,7 +509,7 @@ lazy val app_observe_server_gn_test =
       applicationConfSite := DeploymentSite.GN,
       Universal / mappings := {
         // filter out sjs jar files. otherwise it could generate some conflicts
-        val universalMappings = (mappings in (app_observe_server, Universal)).value
+        val universalMappings = (app_observe_server/Universal/mappings).value
         val filtered          = universalMappings.filter { case (_, name) =>
           !name.contains("_sjs")
         }
@@ -537,7 +537,7 @@ lazy val app_observe_server_gs = preventPublication(project.in(file("app/observe
     applicationConfSite := DeploymentSite.GS,
     Universal / mappings := {
       // filter out sjs jar files. otherwise it could generate some conflicts
-      val universalMappings = (mappings in (app_observe_server, Universal)).value
+      val universalMappings = (app_observe_server/Universal/mappings).value
       val filtered          = universalMappings.filter { case (_, name) =>
         !name.contains("_sjs")
       }
@@ -565,7 +565,7 @@ lazy val app_observe_server_gn = preventPublication(project.in(file("app/observe
     applicationConfSite := DeploymentSite.GN,
     Universal / mappings := {
       // filter out sjs jar files. otherwise it could generate some conflicts
-      val universalMappings = (mappings in (app_observe_server, Universal)).value
+      val universalMappings = (app_observe_server/Universal/mappings).value
       val filtered          = universalMappings.filter { case (_, name) =>
         !name.contains("_sjs")
       }
--- a/build.sbt
+++ b/build.sbt
@@ -335,6 +335,7 @@ lazy val observe_engine = project
 lazy val acm = project
   .in(file("modules/acm"))
   .settings(commonSettings: _*)
+  .enablePlugins(SbtXjcPlugin)
   .settings(
     libraryDependencies ++= Seq(
       EpicsService,
@@ -347,25 +348,8 @@ lazy val acm = project
     ) ++ Logback ++ JAXB,
     Test / libraryDependencies ++= Logback,
     Test / testOptions := Seq(),
-    Compile / sourceGenerators += Def.task {
-      import scala.sys.process._
-      val pkg = "edu.gemini.epics.acm.generated"
-      val log = state.value.log
-      val gen = (Compile / sourceManaged).value
-      val out = pkg.split("\\.").foldLeft(gen)(_ / _)
-      val xsd = sourceDirectory.value / "main" / "resources" / "CaSchema.xsd"
-      val cmd = List("xjc", "-d", gen.getAbsolutePath, "-p", pkg, xsd.getAbsolutePath)
-      val mod = xsd.getParentFile.listFiles.map(_.lastModified).max
-      val cur =
-        if (out.exists && out.listFiles.nonEmpty) out.listFiles.map(_.lastModified).min
-        else Int.MaxValue
-      if (mod > cur) {
-        out.mkdirs
-        val err = cmd.run(ProcessLogger(log.info(_), log.error(_))).exitValue
-        if (err != 0) sys.error("xjc failed")
-      }
-      out.listFiles.toSeq
-    }.taskValue
+    xjcCommandLine ++= Seq("-p", "edu.gemini.epics.acm.generated"),
+    xjcLibs ++= JAXB
   )
 
 /**
--- a/build.sbt
+++ b/build.sbt
@@ -24,6 +24,8 @@ ThisBuild / resolvers += "Gemini Repository".at(
 // Add e.g. a `jres.sbt` file with your particular configuration
 ThisBuild / ocsJreDir := Path.userHome / ".jres8_ocs3"
 
+ThisBuild / evictionErrorLevel := Level.Info
+
 Global / cancelable := true
 
 // Should make CI builds more robust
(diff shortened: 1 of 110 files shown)


cquiroz:
Quick comment, you could remove the packtracker action, we are moving away from it

build.sbt
```diff
@@ -347,25 +351,8 @@ lazy val acm = project
...
-        if (out.exists && out.listFiles.nonEmpty) out.listFiles.map(_.lastModified).min
-        else Int.MaxValue
-      if (mod > cur) {
-        out.mkdirs
-        val err = cmd.run(ProcessLogger(log.info(_), log.error(_))).exitValue
-        if (err != 0) sys.error("xjc failed")
-      }
-      out.listFiles.toSeq
-    }.taskValue
+    xjcCommandLine ++= Seq("-p", "edu.gemini.epics.acm.generated"),
```
cquiroz, line 354: Neat

modules/engine/src/main/scala/observe/engine/Engine.scala
```diff
@@ -460,13 +460,15 @@ class Engine[F[_]: MonadError[*[_], Throwable]: Logger, S, U](stateL: Engine.Sta
     initialState: S,
     f:            (A, S) => F[(S, B, Option[Stream[F, A]])]
   )(implicit ev:  Concurrent[F]): Stream[F, B] =
-    Stream.eval(fs2.concurrent.Queue.unbounded[F, Stream[F, A]]).flatMap { q =>
-      Stream.eval_(q.enqueue1(input)) ++
-        q.dequeue.parJoinUnbounded
+    Stream.eval(cats.effect.std.Queue.unbounded[F, Stream[F, A]]).flatMap { q =>
+      Stream.exec(q.offer(input)) ++
```
cquiroz, line 464: Hadn't seen `exec` before, what does it do?

modules/engine/src/test/scala/observe/engine/StepSpec.scala
```diff
@@ -429,30 +425,31 @@ class StepSpec extends CatsSuite {
...
               .takeThrough(a => !isFinished(a._2.sequences(seqId).status))
               .compile
               .toVector
           )
       }
-      .unsafeRunSync()
 
-    val actionsCompleted = qss.map(_._1).collect { case SystemUpdate(x: Completed[_], _) => x }
-    assert(actionsCompleted.length == 4)
+    qss.map { x =>
```
cquiroz, line 439: I think these are indepedent IO asserts right? If so you need to chain them and return them all at the end or onle the last one will be executed
Like `assert(actionsCompleted.length == 4) *> assert(executionsCompleted...)`
Also you may want to use `assertEquals`

modules/engine/src/test/scala/observe/engine/StepSpec.scala
```diff
@@ -245,21 +236,23 @@ class StepSpec extends CatsSuite {
...
 
-    inside(m.compile.last.unsafeRunSync()) { case Some(Sequence.State.Zipper(zipper, status, _)) =>
-      inside(zipper.focus.toStep) { case Step(_, _, _, _, List(ex1, ex2)) =>
-        assert(
-          Execution(ex1.toList).results.length == 3 && Execution(ex2.toList).actions.length == 1
-        )
+    m.compile.last.map {
+      inside(_) { case Some(Sequence.State.Zipper(zipper, status, _)) =>
+        inside(zipper.focus.toStep) { case Step(_, _, _, _, List(ex1, ex2)) =>
+          assert(
```
cquiroz, line 250: I think here only the last assert is verified

modules/engine/src/test/scala/observe/engine/StepSpec.scala
```diff
@@ -185,7 +179,7 @@ class StepSpec extends CatsSuite {
     case _                       => false
   }
 
-  def runToCompletion(s0: TestState): Option[TestState] =
+  def runToCompletion(s0: TestState): IO[Option[TestState]] =
```
cquiroz, line 182: This name may not be appropriate anymore as we are not really running the IO as before

modules/giapi/src/test/scala/giapi/client/GiapiStatusSpec.scala
```diff
@@ -134,8 +127,7 @@ final class GiapiStatusSpec extends CatsSuite {
         GmpStatus.closeGmpStatus(g) >> c.get[Int](intItemName)
       }
       .attempt
-      .unsafeRunSync() should matchPattern { case Left(GiapiException(_)) =>
-    }
+      .map(matchPattern { case Left(GiapiException(_)) => })
```
cquiroz, line 130: I don't think munit includes this. Is this still on scalatest?

modules/server/src/main/scala/observe/server/ODBSequencesLoader.scala
```diff
@@ -20,7 +19,7 @@ import SeqEvent._
 import ObserveFailure.ObserveException
 import ObserveFailure.UnrecognizedInstrument
 
-final class ODBSequencesLoader[F[_]: ApplicativeError[*[_], Throwable]](
+final class ODBSequencesLoader[F[_]: Async](
```
cquiroz, line 22: Weird that you need this change, What does need `Async`?

build.sbt
```diff
@@ -276,7 +278,8 @@ lazy val observe_server     = project
         TestLibs.value,
         PPrint.value
       ) ++ MUnit.value ++ Http4s ++ Http4sClient ++ PureConfig ++ SeqexecOdb ++ Monocle.value ++ WDBAClient ++
-        Circe.value
+        Circe.value,
+    headerSources / excludeFilter := HiddenFileFilter || (file("modules/server") / "src/main/scala/pureconfig/module/http4s/package.scala").getName
```
rpiaggio, line 282: This is because we are using a copy of the pureconfig-http4s bridge?

modules/server/src/main/scala/observe/server/Systems.scala
```diff
@@ -384,7 +383,7 @@ object Systems {
   }
 
   private def decodeTops(s: String): Map[String, String] =
-    s.split("=|,")
+    s.split("[=,]")
```
cquiroz, line 386: Is this related to xjc change?

modules/server/src/test/scala/observe/server/InstrumentControllerSimSpec.scala
```diff
@@ -51,120 +37,100 @@ class InstrumentControllerSimSpec extends CatsSuite {
...
-      r <- f.join
-    } yield r).unsafeRunSync()
+      _ <- IO.sleep(tick) // give it enough time for at least one tick
+      r <- f.joinWithNever
+    } yield assertEquals(r, ObserveCommandResult.Paused)
 
     // FIXME The simulator should return Stopped
-    // assert(result === ObserveCommandResult.Stopped)
-    assert(result === ObserveCommandResult.Paused)
+    // assertEquals(r, ObserveCommandResult.Stopped)
```
cquiroz, line 108: Commented code

modules/server/src/test/scala/observe/server/tcs/GuideConfigDbSuite.scala
```diff
@@ -151,12 +151,9 @@ final class GuideConfigDbSuite extends munit.FunSuite {
   }
```
cquiroz, line 151: Weird that you can still have `shouldBe` under `munit`

modules/web/server/src/main/scala/observe/web/server/config/package.scala
```diff
@@ -36,49 +34,48 @@ package config {
  */
 package object config {
 
-  implicit val siteReader = ConfigReader.fromCursor[Site] { cf =>
+  implicit val siteReader: AnyRef with ConfigReader[Site] = ConfigReader.fromCursor[Site] { cf =>
```
cquiroz, line 37: Why the `AnyRef`?

modules/web/server/src/main/scala/observe/web/server/logging/AppenderForClients.scala
```diff
@@ -20,7 +20,8 @@ import observe.model.events._
  *
  * This is out of the scala/http4s loop
  */
-class AppenderForClients(out: Topic[IO, ObserveEvent]) extends AppenderBase[ILoggingEvent] {
+class AppenderForClients(out: Topic[IO, ObserveEvent])(implicit ioruntime: IORuntime)
```
cquiroz, line 23: Should we use `Dispatcher` here instead?

modules/web/server/src/test/scala/observe/web/server/http4s/ObserveUIApiRoutesSpec.scala
```diff
@@ -5,36 +5,35 @@ package observe.web.server.http4s
...
     // see https://github.com/http4s/http4s/issues/234
-    val r = (for {
+    for {
       s <- uiRoutes
       r <- Nested(
              s.apply(Request(method = Method.GET, uri = uri("/observe/login"))).value
            ).map(_.status).value
-    } yield r).unsafeRunSync()
-    assert(r === Some(Status.NotFound))
+    } yield assert(r === Some(Status.NotFound))
```
cquiroz, line 32: `assertEquals`

build.sbt
jluhrs: Yes, and I did not want to stamp the AURA license on "borrowed" code.

cquiroz (changes requested):
Great work, This is a big update with many moving parts interacting. I added some comments mostly about tests
Should we copy these changes to the seqexec?

modules/server/src/test/scala/observe/server/InstrumentControllerSimSpec.scala
jluhrs: But, that commented code is relevant to the FIXME

modules/server/src/test/scala/observe/server/tcs/GuideConfigDbSuite.scala
jluhrs: I missed that one, because it compiled nevertheless (because of an exiting import)

modules/web/server/src/main/scala/observe/web/server/config/package.scala
jluhrs: I don't remember writing that. I think it was put when I used an automatic fix for untyped public vars

modules/engine/src/main/scala/observe/engine/Engine.scala
jluhrs: It executes an F[Unit] for its side effect only. In this case, to add an element in the `Queue`

modules/engine/src/test/scala/observe/engine/StepSpec.scala
jluhrs: @cquiroz I may be misunderstanding something, or using the wrong asserts, but those come from munit. They all produce `Unit`, and handle test failures through side effects.

modules/giapi/src/test/scala/giapi/client/GiapiStatusSpec.scala
jluhrs: Yes, the source code imports just that one matcher.

modules/server/src/main/scala/observe/server/Systems.scala
jluhrs: No, it was just a suggestion from the IDE. The expressions are equivalent.

modules/server/src/main/scala/observe/server/ODBSequencesLoader.scala
jluhrs: There are some calls inside the code that require `Temporal`, and others that require `Sync`. `Async` groups both of them.

modules/web/server/src/main/scala/observe/web/server/logging/AppenderForClients.scala
jluhrs: Yes, we should. I'll change it.

cquiroz (approved):
Thanks LGTM

swalker2m:
Will these updates make it into the current seqexec in the interim?

cquiroz:
I think we should to ease maintenance of the seqexec in the interim. @jluhrs We also sue CE in the client, was that ported too?

PR #57, opened 2021-06-17
Ch844 import queries from explore

jluhrs:


modules/common-graphql/src/main/scala/observe/common/ObsQueriesGQL.scala
```diff
@@ -0,0 +1,96 @@
...
+      subscription {
+        observationEdit(programId:"p-2") {
+          id
+        }
+      }
+    """
+  }
+
+  @GraphQL
+  trait ProgramCreateObservation extends GraphQLOperation[ObservationDB] {
```
cquiroz, line 64: Do you need to create/delete observations in `observe`?

project/Settings.scala
```diff
@@ -32,7 +32,7 @@ object Settings {
     val scalaXml            = "1.2.0"
     val catsTime            = "0.3.4"
 
-    val http4sVersion  = "0.23.0-M1"
+    val http4sVersion  = "1.0.0-M21"
```
cquiroz, line 35: Brave

project/plugins.sbt
```diff
@@ -34,4 +34,4 @@ addSbtPlugin("org.scala-sbt.plugins" % "sbt-xjc" % "0.10")
 
 
 
-onLoad in Global := { s => "dependencyUpdates" :: s }
+Global / onLoad := { s => "dependencyUpdates" :: s }
```
cquiroz, line 37: do we still need this in the age of `scala-steward`?

cquiroz (approved):
Looks good
It seems you got the queries from explore. If so, should we share that code into a lib? Or is it going to evolve independently?

modules/common-graphql/src/main/scala/observe/common/ObsQueriesGQL.scala
jluhrs: No

project/Settings.scala
jluhrs: Clue made me do it

jluhrs:
I was talking about the same with Raúl. There some pieces that could be shared, specially the decoders.

> Looks good
> It seems you got the queries from explore. If so, should we share that code into a lib? Or is it going to evolve independently?

PR #81, opened 2021-07-14
Ch991 adapt code to use lucuma observe

jluhrs:
Used Observation.Id and Step.Id from lucuma. Step.Id was the more complicated, because the Id before was an Int that was usually treated as an index.
I left some placeholders that will be filled when coding the actual read of the observation from the GPP ODB.

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/Observation.scala
+++ b/modules/model/shared/src/main/scala/observe/model/Observation.scala
@@ -5,40 +5,49 @@ package observe.model
 
 import cats.Order
 import cats.Show
+import cats.Eq
 import lucuma.core.math.Index
 
 object Observation {
 
+  type Id = lucuma.core.model.Observation.Id
+
   /** An observation is identified by its program and a serial index. */
-  final case class Id(pid: Program.Id, index: Index) {
+  final case class Name(pid: Program.Id, index: Index) {
     def format: String =
       s"${ProgramId.fromString.reverseGet(pid)}-${Index.fromString.reverseGet(index)}"
   }
-  object Id {
+  object Name {
 
-    def fromString(s: String): Option[Observation.Id] =
+    def fromString(s: String): Option[Observation.Name] =
       s.lastIndexOf('-') match {
         case -1 => None
         case n  =>
           val (a, b) = s.splitAt(n)
           Index.fromString.getOption(b.drop(1)).flatMap { i =>
-            Program.Id.fromString.getOption(a).map(Observation.Id(_, i))
+            Program.Id.fromString.getOption(a).map(Observation.Name(_, i))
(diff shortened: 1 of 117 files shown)


modules/model/shared/src/main/scala/observe/model/Observation.scala
```diff
@@ -5,40 +5,49 @@ package observe.model
 
 import cats.Order
 import cats.Show
+import cats.kernel.Eq
```
cquiroz, line 8: `cats.Eq`

modules/model/shared/src/main/scala/observe/model/StepState.scala
```diff
@@ -31,10 +31,10 @@ object StepState {
     }
 
   implicit class StepStateOps(val s: StepState) extends AnyVal {
-    def canSetBreakpoint(i: Int, firstRunnable: Int): Boolean = s match {
+    def canSetBreakpoint: Boolean = s match {
       case StepState.Pending | StepState.Skipped | StepState.Paused | StepState.Running |
           StepState.Aborted =>
-        i > firstRunnable
+        true
```
cquiroz, line 37: Is it so that you can always set a breakpoint?

modules/server/src/main/scala/observe/server/ObserveEnvironment.scala
```diff
@@ -12,17 +12,17 @@ import observe.server.tcs.Tcs
...
-  config:   CleanConfig,
-  stepType: StepType,
-  obsId:    Observation.Id,
-  dataId:   DataId,
-  inst:     InstrumentSystem[F],
-  insSpecs: InstrumentSpecifics,
-  otherSys: List[System[F]],
-  headers:  HeaderExtraData => List[Header[F]],
-  ctx:      HeaderExtraData
+  odb:       OdbProxy[F],
```
cquiroz, line 15: Do we need an `OdbProxy`?

cquiroz (approved):
I didn't review the test part, but the rest seems fine.

modules/model/shared/src/main/scala/observe/model/StepState.scala
jluhrs: No, I changed the logic a bit, because the `i > firstRunnable` part does not make sense with `StepId`. So now the function in `StepState` only checks that the state is right for setting a breakpoint. The caller then checks that the step is after the `firstRunnable` in the list order.

modules/server/src/main/scala/observe/server/ObserveEnvironment.scala
jluhrs: Yes, but with a different implementation.

PR #89, opened 2021-07-26
Synchronized the schema classes with explore.

jluhrs:


rpiaggio (approved):
LGTM

jluhrs:
@toddburnside Yes! Some of them are already, but I need to port the code to use the latest version of lucuma to use them.

PR #147, opened 2021-09-21
Ch1039 update lucuma version

jluhrs:
Adapted the code to use lucuma-core 0.13.2

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/ModelLenses.scala
+++ b/modules/model/shared/src/main/scala/observe/model/ModelLenses.scala
@@ -10,14 +10,12 @@ import lucuma.core.math.Offset
 import lucuma.core.optics.Format
 import lucuma.core.syntax.all._
 import monocle._
-import monocle.function.At.at
 import monocle.function.At.atMap
-import monocle.function.FilterIndex.filterIndex
+import monocle.function.FilterIndex
 import monocle.macros.GenLens
 import monocle.macros.GenPrism
 import monocle.std.option.some
 import monocle.std.string._
-import monocle.unsafe.MapTraversal._
 import observe.model.enum._
 import observe.model.events._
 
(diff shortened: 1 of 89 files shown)


modules/model/shared/src/main/scala/observe/model/ModelLenses.scala
```diff
@@ -99,31 +108,55 @@ trait ModelLenses {
...
+      .andThen( // Find the sequence view
+        sessionQueueL
+      )
+      .andThen( // Find the queue
+        eachViewT
+      )
+      .andThen( // each sequence on the queue
+        obsStepsL
+      )
+      .andThen( // sequence steps
```
cquiroz, line 150: Also these may be written as `lens.each` instead of `andThen(each)`

modules/model/shared/src/main/scala/observe/model/SequenceView.scala
```diff
@@ -39,5 +39,5 @@ object SequenceView {
     Eq.by(x => (x.idName, x.metadata, x.status, x.steps, x.willStopIn))
 
   val stepT: Traversal[SequenceView, Step] =
-    SequenceView.steps ^|->> each
+    SequenceView.steps.andThen(each[List[Step], Step])
```
cquiroz, line 42: `SequenceView.steps.each`?

modules/web/client/src/main/scala/observe/web/client/components/LogArea.scala
```diff
@@ -158,7 +158,7 @@ object LogArea {
   object State {
 
     def levelLens(l: ServerLogLevel): Lens[State, Option[Boolean]] =
-      State.selectedLevels ^|-> at(l)
+      State.selectedLevels.andThen(at(l))
```
cquiroz, line 161: Maybe `at` can also be inlined? `State.selectedLevel.at(l)`?

cquiroz (approved):
This looks great, it is a big change. I think is some cases the lens compositions could be simplified but it is low priority

cquiroz:
Could we get this ported to the seqexec? Pretty please 🙏

PR #161, opened 2021-09-30
Use shared GraphQL schema (lucuma-schemas).

rpiaggio:


jluhrs (approved):
LGTM, thanks fo the change.

PR #165, opened 2021-10-05
Remove unnecessary classes for GQL queries.

rpiaggio:
The removed classes are used in explore mainly to enable passing results from different queries into the same components.

 Here that doesn't seem necessary and data can be accessed directly through the types generated by clue.


jluhrs (approved):
LGTM

PR #291, opened 2022-01-28
Adopt sbt-lucuma 0.9.4

armanbilge:
Opening for CI.


armanbilge:
Ok, CI failed because it looks like clue is publishing JDK 11 artifacts. So should I bump up to JDK 11 here (or upstream in sbt-lucuma), or bump down to JDK8 in clue?

build.sbt
```diff
@@ -51,30 +46,6 @@ Global / concurrentRestrictions += Tags.limit(ScalaJSTags.Link, 2)
...
-    ""
-  }
-  s"-${out.commitSuffix.sha}$dirtySuffix"
-}
-
-def fallbackVersion(d: java.util.Date): String = s"HEAD-${sbtdynver.DynVer.timestamp(d)}"
-
-val dateFormatter = java.time.format.DateTimeFormatter.BASIC_ISO_DATE
-
-inThisBuild(
```
cquiroz, line 68: In applications which we won't release as libraries we have been using this version format as it is displayed to the end users

I'm not sure if we want to change it, or maybe it should be push down to the part where the actual releases are made

build.sbt
```diff
@@ -124,7 +104,6 @@ lazy val giapi = project
   .enablePlugins(GitBranchPrompt)
   .settings(commonSettings: _*)
   .settings(
-    addCompilerPlugin(Plugins.kindProjectorPlugin),
```
cquiroz, line 127: Why don't we need this anymore?

cquiroz (changes requested):
I think we want to keep the version scheme as it is now as this is not a library and the version is referenced in the code

build.sbt
armanbilge: You get it for free from sbt-typelevel-settings.

build.sbt
armanbilge: Huh. So the version of your published artifacts doesn't match your tags?

cquiroz: Observe and explore are end user applications, we don't publish them as artifacts but as full apps

In the case of explore it is just web assets, js, css, etc

Observe is released as a tarball

armanbilge:
I see, clue requires JDK 11 because of the jdk-http-client which then pushes JDK 11 on downstreams. Thoughts?

armanbilge:
Hmm, the JS job seems to be stuck 😕 I blame the scoverage.

armanbilge:
Yup that seems to have fixed it 😆

modules/web/client/src/main/scala/observe/web/client/ObserveApp.scala
```diff
@@ -28,7 +27,7 @@ final class ObserveLauncher[F[_]](implicit val F: Sync[F], L: LiftIO[F]) {
   def serverSite: F[Site] =
     L.liftIO(IO.fromFuture {
       IO {
-        import ExecutionContext.Implicits.global
+        import org.scalajs.macrotaskexecutor.MacrotaskExecutor.Implicits._
```
cquiroz, line 30: Cool, this should enable us to use scala.js 1.8

armanbilge: Oh, this PR upgraded you already 😆 I didn't realize

.scalafix.conf
```diff
@@ -1 +1,5 @@
+include ".scalafix-common.conf"
+
+rules = [] # Disable organize imports for now
```
armanbilge, line 3: The organize imports was breaking the tests with ambiguous implicits for arbitraries. I guess some of them are sensitive to the import order which is a bit strange.

armanbilge:
@jluhrs why did you close this PR? :)

armanbilge:
Oh, it looks like it was merged as part of https://github.com/gemini-hlsw/observe/pull/489. Sorry I missed that!

armanbilge:
Wait, no, sorry, I'm very confused 😂 this PR is still good for merging, I hope.

PR #485, opened 2022-08-04
rollback to http4s 0.23.x

armanbilge:
- [x] rollback to 0.23 series
- [x] schism http4s module versions
- [x] address deprecatia
  - [x] swap jdk-http-client for async-http-client

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -34,7 +34,7 @@ object Settings {
 
     val http4sVersion                  = "0.23.14-27-fd0137d-SNAPSHOT"
     val http4sBlazeVersion             = "0.23.12"
-    val http4sAsyncHttpClientVersion   = "0.23.11"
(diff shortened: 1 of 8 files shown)


project/Settings.scala
```diff
@@ -32,7 +32,13 @@ object Settings {
     val scalaXml            = "1.2.0"
     val catsTime            = "0.4.0"
 
-    val http4sVersion  = "1.0.0-M27"
+    val http4sVersion                  = "0.23.14-27-fd0137d-SNAPSHOT"
```
armanbilge, line 35: We needed the snapshot for an unreleased feature, we can hold this PR until the next release.

jluhrs: Because Observe is a work in progress, a long way away for being in production, I say it is acceptable to have it depend on a snapshot temporarily.

PR #655, opened 2023-06-19
Migrate to latest version of clue

rpiaggio:
The `.applyP(client)` are not necessary in Scala 3 with an implicit client, but in Scala 2 all the necessary implicit combinations do not seem to kick in. I'll check what I can do in clue to improve this, but in the meantime this works.


cquiroz (approved):
LGTM

PR #657, opened 2023-06-20
New web

rpiaggio:
The new web is still an isolated demo, but since we are going to start evolving it into a functional web, I think it's a good idea to merge so as not to diverge even more.

PR #666, opened 2023-06-21
Update versions

rpiaggio:
Updates to latest versions of `lucuma-core`, `lucuma-schemas` and `lucuma-ui`.

Some code adaptations were required.

This PR subsumes https://github.com/gemini-hlsw/observe/pull/657

### DIFF
--- a/modules/server_new/src/clue/scala/observe/common/ObsQueriesGQL.scala
+++ b/modules/server_new/src/clue/scala/observe/common/ObsQueriesGQL.scala
@@ -13,13 +13,15 @@ import lucuma.core.math
 import lucuma.core.enums
 import lucuma.core.model
 import cats.syntax.functor.*
-import lucuma.core.model.sequence.{Atom, ExecutionSequence, GmosGratingConfig, StaticConfig, Step}
+import lucuma.core.model.sequence.{Atom, ExecutionSequence, Step}
+import lucuma.core.model.sequence.gmos.{DynamicConfig, GmosGratingConfig, StaticConfig}
 
 import java.time
 import lucuma.core.model.{ExecutionEvent, Observation, Target}
 
-// gql: import lucuma.schemas.decoders._
-// gql: import io.circe.refined._
+// gql: import lucuma.schemas.decoders.given
+// gql: import io.circe.refined.*
+// gql: import lucuma.odb.json.sequence.given
 
 object ObsQueriesGQL {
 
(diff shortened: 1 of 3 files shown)


modules/new_model/shared/src/main/scala/observe/model/Conditions.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package observe.model
+
+import cats.Eq
+import cats.derived.*
+import observe.model.enums.*
+
+final case class Conditions(
+  cc: CloudCover,
+  iq: ImageQuality,
```
cquiroz, line 12: We have new classes (not enumerated) classes for conditions like `CloudExtinction`. Will we add those later on?

modules/new_model/shared/src/main/scala/observe/model/enums/SystemName.scala
```diff
@@ -0,0 +1,29 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package observe.model.enums
+
+import cats.Eq
+import cats.derived.*
+import cats.syntax.eq.*
+import observe.model.ParamName
+
+enum SystemName(val system: String) derives Eq:
```
cquiroz, line 11: Why not `derives Enumerated`?

modules/web/src/main/scala/observe/ui/Icons.scala
```diff
@@ -0,0 +1,168 @@
...
+package observe.ui
+
+import react.fa.FAIcon
+import react.fa.FontAwesomeIcon
+import react.fa.IconLibrary
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.*
+
+object Icons:
```
cquiroz, line 13: I wonder is some of these could go to `lucuma-ui` Maybe it just make things complicated?

modules/new_model/shared/src/main/scala/observe/model/Conditions.scala
rpiaggio: Yes, we should do that.

modules/new_model/shared/src/main/scala/observe/model/enums/SystemName.scala
rpiaggio: Good idea, I'll implement and remove the parser.

modules/web/src/main/scala/observe/ui/Icons.scala
rpiaggio: IDK what's best here, whether to have everything in one place or each system with their own references for the icons they use.

PR #688, opened 2023-07-05
Query ready observations from UI client and versions update

rpiaggio:
This PR:
- Brings dependencies up to date, including `sbt-lucuma`, which introduces linting.
- Removes unused code from web project.
- Makes web project query observations with status `READY` from ODB. Sequence is not queried yet, it is still static.
  - While we don't define an authorization mechanism for the ODB UI, I'm using temporary tokens to connect to the ODB (obtained from logging in from Explore). In other words, if you try to run the UI, it will not work unless the token is updated (it's hardcoded in `Main.scala`).


rpiaggio:
CI is failing now because of linting of unused code outside of the web project. I started cleaning it up but it seems that we want to keep some of the code for the moment. @jluhrs can you take a look and decide what we can delete and what we want to keep commented for now?

rpiaggio:
I disabled fatal warnings for the moment so we can move forward while still migrating.

jluhrs (approved):
LGTM

PR #706, opened 2023-07-25
Login

rpiaggio:
This PR implements login for observe web.

For the moment, login is required to show anything and a staff role is required too. We can change this in the future, and allow access to a readonly version for non-logged or non-staff users.

A token is still required to access the ODB though, so it's a pending discussion whether we will:
- use an exposed service token for readonly access; or
- allow certain queries in the ODB without a security token; or
- require all users to log into ODB, even to access it readonly.

### DIFF
--- a/modules/web/src/main/scala/observe/ui/Main.scala
+++ b/modules/web/src/main/scala/observe/ui/Main.scala
@@ -5,24 +5,28 @@ package observe.ui
 
 import cats.effect.IO
 import cats.effect.Resource
-import cats.effect.Sync
 import cats.effect.std.Dispatcher
 import cats.effect.unsafe.implicits._
 import cats.syntax.all.*
 import clue.js.WebSocketJSBackend
 import clue.js.WebSocketJSClient
-import clue.websocket.CloseParams
 import clue.websocket.ReconnectionStrategy
+import crystal.react.*
 import crystal.react.hooks.*
-import io.circe.Json
-import io.circe.syntax.given
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.extra.router._
 import japgolly.scalajs.react.vdom.html_<^.*
 import log4cats.loglevel.LogLevelLogger
+import lucuma.core.model.StandardRole
+import lucuma.core.model.StandardUser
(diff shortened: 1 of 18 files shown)


rpiaggio:
BTW, for development, observe should now be accessed as `https://local.lucuma.xyz:8081`.

modules/web/src/main/scala/observe/ui/Main.scala
```diff
@@ -5,67 +5,85 @@ package observe.ui
...
+  private val (router, _ /*routerCtl*/ ) =
     RouterWithProps.componentAndCtl(BaseUrl.fromWindowOrigin, Routing.config)
 
-  val mainApp =
+  private val mainApp =
     ScalaFnComponent
-      .withHooks[Unit]
-      .useStateView(RootModel.Initial)
-      .render((_, state) => router(state))
+      .withHooks[Either[Throwable, Option[UserVault]]]
```
cquiroz, line 63: No `Props`?

modules/web/src/main/scala/observe/ui/components/Home.scala
```diff
@@ -85,81 +86,94 @@ object Home {
...
-                    props.rootModel.get.status,
-                    props.rootModel.get.operator,
-                    props.rootModel.zoom(RootModel.conditions)
-                  )
-                )
-              )
+      .render: (props, ctx, observations, demo) =>
+        import ctx.given
+
+        props.rootModel.get.userVault.map(userVault =>
```
cquiroz, line 92: Are we going to have side tabs like in explore?

cquiroz (approved):
LGTM

modules/web/src/main/scala/observe/ui/Main.scala
rpiaggio: It's such a small component that it didn't seem necessary, but we can extract it into another file with a `case class` for props if need be.

modules/web/src/main/scala/observe/ui/components/Home.scala
rpiaggio: I hadn't thought of that. Do you think it would be useful?

PR #719, opened 2023-08-04
Replace old sp model with new ODB model (partial implementation)

jluhrs:


modules/engine/src/main/scala/observe/engine/Sequence.scala
```diff
@@ -10,25 +10,33 @@ import observe.engine.Action.ActionState
 import observe.engine.Result.RetVal
 import observe.model.SequenceState
 import lucuma.core.model.Observation
-import lucuma.core.model.sequence.Step.{Id => StepId}
+import lucuma.core.model.sequence.Atom as OcsAtom
```
rpiaggio, line 13: Is the alias needed?

modules/engine/src/main/scala/observe/engine/Sequence.scala
```diff
@@ -10,25 +10,33 @@ import observe.engine.Action.ActionState
...
 
 /**
  * A list of `Step`s grouped by target and instrument.
  */
-final case class Sequence[F[_]](
-  id:    Observation.Id,
-  steps: List[Step[F]]
+final case class Sequence[F[_]] private (
+  id:     Observation.Id,
+  atomId: Option[OcsAtom.Id],
```
rpiaggio, line 21: The `Sequence` is a flattened list of `Steps`? Why is the `Atom` optional?

rpiaggio (approved):
LGTM. TBH, I haven't looked in depth at all the changes.

modules/engine/src/main/scala/observe/engine/Sequence.scala
jluhrs: To allow for the special case of an empty Sequence. For example, one that has run to completion.

PR #752, opened 2023-09-05
Breakpoints, share sequence columns with explore, sequence table styling

rpiaggio:
- Toggable breakpoints (initialized from ODB).
- All the state that will come from the observe server is concentrated in the `ExecutionState` class.
- All possible code is now shared with explore.
- Execution state is shown in a new subrow, freeing up horizontal space.
- Acquisition and science stables separated.
- Styling follows inception spec.

Dark mode:
image

Light mode:
image

Coming up next: 
- Atom boundaries.
- The black text in the `Gcal` button is the standard `Warning` severity button of prime react. We should fix that in the style.


cquiroz (approved):
LGTM

PR #774, opened 2023-09-21
Test porting

cquiroz:
This PR includes some general cleanups and ports of a few tests:

* Remove ScalaTest, use only munit
* Some build cleanups
* Use cats-time rather than `io.chrisdavenport.time`
* Remove many warnings (inline Focus)


modules/engine/src/test/scala/observe/engine/SequenceSuite.scala
```diff
@@ -208,7 +204,7 @@ class SequenceSpec extends AnyFlatSpec {
   val seqzar0: Sequence.Zipper[IO]                                      = simpleSequenceZipper(stepzar0)
   val seqzar1: Sequence.Zipper[IO]                                      = simpleSequenceZipper(stepzar1)
 
-  "next" should "be None when there are no more pending executions" in {
+  test("be None when there are no more pending executions") {
```
jluhrs, line 207: In this test and the following, we lose a bit of information in the test description by omitting the name of the tested function.

modules/server_new/src/main/scala/observe/server/SequenceData.scala
```diff
@@ -28,4 +27,6 @@ object SequenceData {
 
   def seq[F[_]]: Lens[SequenceData[F], Sequence.State[F]] = Focus[SequenceData[F]](_.seq)
 
+  def visitId[F[_]] = Focus[SequenceData[F]](_.visitId)
```
jluhrs, line 30: Public members should have an explicit type.

jluhrs (approved):
LGTM

PR #785, opened 2023-09-26
Atom boundaries

rpiaggio:
This implied a general table row separator rework, where we stop using `border` in favor of `box-shadow` to paint row separators, hover indicator, breakpoints and atom boundaries.


cquiroz (approved):
LGTM

PR #788, opened 2023-09-28
Web server

cquiroz:
This starts porting the old web server to let the UI connect.
There are some important differences from the original server for the seqexec:

* No boopickle, messages are sent using json
* No auth, it will be added later
* Only a few endpoints are supported (changing conditions)
* No serving of static files

Anyway this should let the UI work to continue and we can build on top of this the rest of the requirements


jluhrs (approved):
LGTM

PR #789, opened 2023-09-29
Restore sequence load tests

jluhrs:


modules/model/shared/src/main/scala/observe/model/enums/Resource.scala
```diff
@@ -44,24 +44,26 @@ sealed abstract class Instrument(tag: String, ordinal: Int, label: String)
...
-  case object Gnirs extends Instrument("Gnirs", 15, "GNIRS")
-  case object Gpi   extends Instrument("Gpi", 16, "GPI")
-  case object Gsaoi extends Instrument("Gsaoi", 17, "GSAOI")
-  case object Niri  extends Instrument("Niri", 18, "NIRI")
-  case object Nifs  extends Instrument("Nifs", 19, "NIFS")
+//  case object Gnirs extends Instrument("Gnirs", 15, "GNIRS")
+//  case object Gpi   extends Instrument("Gpi", 16, "GPI")
+//  case object Gsaoi extends Instrument("Gsaoi", 17, "GSAOI")
+//  case object Niri  extends Instrument("Niri", 18, "NIRI")
+//  case object Nifs  extends Instrument("Nifs", 19, "NIFS")
```
jluhrs, line 55: I removed the unsupported instrument to remove the need to deal with them in match statements. They must be restored as support for each instrument is put back in.

modules/server_new/src/main/scala/observe/server/InstrumentSystem.scala
```diff
@@ -39,17 +39,17 @@ trait InstrumentSystem[F[_]] extends System[F] {
 object InstrumentSystem {
   val ObserveOperationsTimeout = 1.minute
 
-  final case class StopObserveCmd[F[_]](self: Boolean => F[Unit])
-  final case class AbortObserveCmd[F[_]](self: F[Unit])
-  final case class PauseObserveCmd[F[_]](self: Boolean => F[Unit])
+  final case class StopObserveCmd[+F[_]](self: Boolean => F[Unit])
```
jluhrs, line 42: This avoids the warning  "type test cannot be checked at runtime" in match statements.

modules/server_new/src/main/scala/observe/server/ODBSequencesLoader.scala
```diff
@@ -93,18 +93,17 @@ object ODBSequencesLoader {
...
+    l:        Lens[EngineState[F], Option[SequenceData[F]]]
   ): Endo[EngineState[F]] = st =>
     l.replace(
       SequenceData[F](
         observer,
         none,
         SystemOverrides.AllEnabled,
         seqg,
-        execEngine.load(
+        Engine.load(
```
jluhrs, line 106: The method didn't use any of the instance state, so I moved it to the companion object. That removes the need to pass an instance of `Engine` around.

modules/model/shared/src/main/scala/observe/model/enums/Resource.scala
cquiroz: Why not use the lucuma core Instrument?

modules/server_new/src/test/scala/observe/server/TestCommon.scala
```diff
@@ -3,29 +3,77 @@
...
+import cats.data.NonEmptyList
 import cats.effect.IO
 import cats.syntax.all.*
-import cats.Applicative
-import lucuma.core.enums.Site
+import cats.{Applicative, Monoid}
+import eu.timepit.refined.types.numeric.PosLong
+//import giapi.client.ghost.GhostClient
+//import giapi.client.gpi.GpiClient
+import lucuma.core.enums.{
```
cquiroz, line 13: maybe `lucuma.core.enums.*`?

PR #793, opened 2023-09-29
Restored the watchdog of the main stream process.

jluhrs:
The implementation is a bit of a hack. It uses an existing event to join the heartbeat stream with the main event stream.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
+++ b/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
@@ -841,14 +841,15 @@ object ObserveEngine {
     }
 
     override def eventStream: Stream[F, ObserveEvent] =
-      stream(EngineState.default[F]).flatMap(x => Stream.eval(notifyODB(x).attempt)).flatMap {
-        case Right((ev, qState)) =>
-          val sequences = List(qState.selected.gmosNorth, qState.selected.gmosSouth).flattenOption
-            .map(viewSequence)
-          toObserveEvent[F](ev, qState)
-        case Left(x)             =>
-          Stream.eval(Logger[F].error(x)("Error notifying the ODB").as(NullEvent))
-      }
+      Stream.eval(executeEngine.offer(Event.getState(_ => heartbeatStream.some)).as(NullEvent)) ++
+        stream(EngineState.default[F]).flatMap(x => Stream.eval(notifyODB(x).attempt)).flatMap {
+          case Right((ev, qState)) =>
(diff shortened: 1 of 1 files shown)


modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -841,14 +841,15 @@ object ObserveEngine {
...
-          val sequences = List(qState.selected.gmosNorth, qState.selected.gmosSouth).flattenOption
-            .map(viewSequence)
-          toObserveEvent[F](ev, qState)
-        case Left(x)             =>
-          Stream.eval(Logger[F].error(x)("Error notifying the ODB").as(NullEvent))
-      }
+      Stream.eval(executeEngine.offer(Event.getState(_ => heartbeatStream.some)).as(NullEvent)) ++
+        stream(EngineState.default[F]).flatMap(x => Stream.eval(notifyODB(x).attempt)).flatMap {
+          case Right((ev, qState)) =>
+            val sequences = List(qState.selected.gmosNorth, qState.selected.gmosSouth).flattenOption
```
cquiroz, line 847: Is `sequences` being used?

modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -841,14 +841,15 @@ object ObserveEngine {
...
     override def eventStream: Stream[F, ObserveEvent] =
-      stream(EngineState.default[F]).flatMap(x => Stream.eval(notifyODB(x).attempt)).flatMap {
-        case Right((ev, qState)) =>
-          val sequences = List(qState.selected.gmosNorth, qState.selected.gmosSouth).flattenOption
-            .map(viewSequence)
-          toObserveEvent[F](ev, qState)
-        case Left(x)             =>
-          Stream.eval(Logger[F].error(x)("Error notifying the ODB").as(NullEvent))
-      }
+      Stream.eval(executeEngine.offer(Event.getState(_ => heartbeatStream.some)).as(NullEvent)) ++
```
cquiroz, line 844: Is `heartbeat` emiting an event and we detect when it doesn't happen?

modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
jluhrs: No. That was a leftover of the Metrics.

cquiroz (approved):
LGTM

PR #794, opened 2023-09-29
Send the initial state on the first connection

cquiroz:
Upon connection a client will get two messages. The first is some metadata like client id and then a message with the current state

The ability to send messages to a single client is now restored:

```json
{"site":"GS","clientId":"edfb83fd-4582-4df4-9678-f4f264e876ee","version":"20230929-0bccfdac-UNCOMMITED"}
{"conditions":{"cc":null,"iq":null,"sb":null,"wv":null}}
```


jluhrs (approved):
LGTM

PR #798, opened 2023-10-02
Unify shared model + TimeSpan migration + UI http client

rpiaggio:
When migrating to Scala 3 and the new model, we did diverging shared model migrations for the UI and server. This PR started as an attempt to unify the changes so that we have a unified shared model again. Some of the changes propagated further than the shared module though.

Among the changes:

- Switched some ADTs to `enum`, where possible.

- Unified NodAndShuffle prefix to `Ns`.

- `Progress` conflicts with our progress bar component. Renamed to `ObservationProgress`. Also changed `FiniteDuration` -> `TimeSpan`. `FiniteDuration` is not meant to model things. `TimeSpan` further ensures that it's a positive duration. I took the liberty to propagate this change into server model classes up to where we do need `FiniteDuration` to interact with concurrency APIs. This ended up being a large part of this PR, sorry if it's too invasive. If there are good reasons to use `FiniteDuration` instead, we can switch back.

Some question moving further:
- Can we drop `observe.model.Step` and switch to `lucuma.core.model.sequence.Step`?

- Finally: Is there a good reason to keep the `Instrument` enumeration in `Resource.scala`? Could we define a `Resource` that wraps `lucuma.core.enums.Instrument` instead? We would have to address the fact that observe does not support all the instruments from core. Also, it's not clear to me whether the `ordinal` can be changed or is it used to interact with other systems?

Also in this PR: incipient logic to start connecting the client to the server, basically creating an HTTP client and interface.


modules/model/shared/src/main/scala/observe/model/ModelLenses.scala
```diff
@@ -38,10 +36,10 @@ trait ModelLenses {
 //  val parametersRoot: Iso[Map[ParamName, ParamValue], Map[ParamName, ParamValue]]   =
 //    Iso.id[Map[ParamName, ParamValue]]
 
-  val sequenceStepT: Traversal[SequenceView, Step] =
```
rpiaggio, line 41: It seems we will no longer have sequences in the shared or even the server model. I was thinking of removing these `Traversal`s.

modules/model/shared/src/test/scala/observe/model/arb/ArbStepConfig.scala
```diff
@@ -23,14 +21,14 @@ trait ArbStepConfig {
       b <- asciiStr
     } yield (a, b)
 
-  val parametersGen: Gen[Parameters] =
-    Gen.chooseNum(0, 10).flatMap(s => Gen.mapOfN[String, String](s, stepItemG))
+  // val parametersGen: Gen[Parameters] =
```
rpiaggio, line 24: We are removing `Parameters`, right?

cquiroz:
I support having a single `Instrument` enum

modules/model/shared/src/main/scala/observe/model/ModelLenses.scala
cquiroz: 👍

rpiaggio:
> I support having a single `Instrument` enum

I was thinking we can have a `Resource` instance that wraps a `lucuma.core.enums.Instrument` and a `Resource.fromInstrument: Option[Resource]` constructor that checks if it's a valid observe instrument.

modules/model/shared/src/main/scala/observe/model/ObservationProgress.scala
```diff
@@ -4,88 +4,42 @@
...
+    obsId:     Observation.Id,
+    stepId:    Step.Id,
+    total:     TimeSpan,
+    remaining: TimeSpan,
+    stage:     ObserveStage,
+    sub:       NsSubexposure
+  ) extends ObservationProgress(true)
+
+object ObservationProgress:
+  implicit val regular: Prism[ObservationProgress, ObservationProgress.Regular] =
```
cquiroz, line 41: An `implicit Prism`?

rpiaggio: Don't know how that got there. Removed.

modules/model/shared/src/main/scala/observe/model/enums/NSAction.scala
```diff
@@ -5,19 +5,10 @@ package observe.model.enums
...
-  case object StageObserveComplete extends NSAction("StageObserveComplete")
-  case object Done                 extends NSAction("Done")
-
-  /** @group Typeclass Instances */
-  given Enumerated[NSAction] =
-    Enumerated
-      .from(Start, NodStart, NodComplete, StageObserveStart, StageObserveComplete, Done)
-      .withTag(_.tag)
-}
+enum NsAction(val tag: String) derives Enumerated:
```
cquiroz, line 8: I think `NSAction` is correct. It stands for `Nod & Shuffle Action`

modules/model/shared/src/main/scala/observe/model/operations.scala
```diff
@@ -76,92 +68,78 @@ object operations {
               )
             }
           }
-        case NsCycleLevel     =>
+        case NsCycle     =>
```
cquiroz, line 71: I'd say this should `NSCycle` too

modules/model/shared/src/main/scala/observe/model/operations.scala
```diff
@@ -76,92 +68,78 @@ object operations {
...
           List(Operations.PauseGracefullyObservation, Operations.StopGracefullyObservation)
-        case NsNodLevel       =>
+        case NsNod       =>
```
cquiroz, line 73: `NSNod`

modules/model/shared/src/test/scala/observe/model/arb/ArbNSRunningState.scala
```diff
@@ -5,25 +5,25 @@ package observe.model.arb
 
 import lucuma.core.util.arb.ArbEnumerated.*
 import observe.model.*
-import observe.model.arb.ArbNSSubexposure.given
+import observe.model.arb.ArbNsSubexposure.given
```
cquiroz, line 8: `ArbNSSubexposure`

modules/model/shared/src/test/scala/observe/model/arb/ArbStepConfig.scala
cquiroz: 👍

modules/server_new/src/main/scala/observe/server/gmos/GmosControllerSim.scala
```diff
@@ -25,21 +27,18 @@ import observe.server.ObsProgress
...
 import org.typelevel.log4cats.Logger
 
-import scala.concurrent.duration.Duration
-import scala.concurrent.duration.FiniteDuration
-
 /**
  * Keep track of the current execution state
  */
-final case class NSCurrent(
+final case class NsCurrent(
```
cquiroz, line 37: I'd keep `NSCurrent`

cquiroz (commented):
Some comments inline but it seems very good

cquiroz (approved):
LGTM

modules/server_new/src/main/scala/observe/server/gmos/GmosControllerSim.scala
rpiaggio: We had some `NS` and some `Ns`. To unify, I took the approach we usually see in other acronyms: `Http`, `Uri`, `Js`, etc.

rpiaggio: We use that convention for instruments too: `Gmos`, `Gnirs`, `Gsaoi`, etc.

cquiroz: Ok, that seems fine

PR #800, opened 2023-10-03
Switch to shared giapi-scala lib

armanbilge:
https://github.com/gemini-hlsw/giapi-scala


project/Settings.scala
```diff
@@ -246,6 +247,7 @@ object Settings {
     )
 
     // GIAPI Libraries
+    val GiapiScala         = "edu.gemini"      %% "giapi"         % LibraryVersions.giapiScala
```
cquiroz, line 250: I bet we can removee some more dependencies, like activemq

cquiroz (approved):
Looks good to me, I think we can simplify the build removing giapi dependencies

PR #804, opened 2023-10-03
Move web client to subfolder

rpiaggio:
Puts the UI client back in `modules/web/client` (instead of `modules/web`).


cquiroz (approved):
LGTM

PR #805, opened 2023-10-03
Bump lucuma-ui (removes broadcast lib), add running step to ExecutionState

rpiaggio:


modules/web/client/src/main/scala/observe/ui/BroadcastEvent.scala
```diff
@@ -3,16 +3,20 @@
...
 package observe.ui
 
+import fs2.dom.Serializer
+
 import scala.scalajs.js
 
 // These are messages sent across tabs thus they need to be JS compatible
 // We don't need yet more than just an index to differentiate
-sealed trait BroadcastEvent extends js.Object {
+sealed trait BroadcastEvent extends js.Object:
```
cquiroz, line 12: Are we also logging out of all tabs?
I wonder if this will work across apps, if you logout  of observe, you logout of eexplore

cquiroz (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/BroadcastEvent.scala
rpiaggio: Yup, it was easier just to use the same behavior in both explore and observe.

I don't think logging out of one logs you out of the other. They have different channel names. But we could unify if that's the desired behavior.

PR #806, opened 2023-10-03
Configuration tab, conditions being sent to server

rpiaggio:
This PR sets up a proxy to communicate with the web server in dev mode.

Furthermore, changing conditions is actually invoking the server.


modules/web/client/src/main/publicdev/environments.conf.json
```diff
@@ -2,7 +2,7 @@
   {
     "hostName": "*",
     "environment": "DEVELOPMENT",
-    "odbURI": "wss://lucuma-postgres-odb-dev.herokuapp.com/ws",
+    "odbURI": "ws://localhost:8082/ws",
```
cquiroz, line 5: Do we need to run a local. odb?

cquiroz (approved):
LGTM, though I'm not sure if I want to have the default odb as localhost

modules/web/client/src/main/publicdev/environments.conf.json
rpiaggio: This slipped through, I'll revert. We have logic in place to pick up `local.conf.json` if need be.

rpiaggio: Actually, this shouldn't be checked in, it got out of sync with `.gitignore` with the directory change.

PR #808, opened 2023-10-04
Refactor initialization, set up WS connection

rpiaggio:
I moved all initialization to React. This way we can show a loader as soon as possible while we initialize. Also, we use the available hooks.

Added to the initialization is establishing the WS connection to the server and processing the received events.

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
@@ -1,0 +1,216 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package observe.ui.components
+
+import cats.syntax.all.*
+import cats.effect.IO
+import cats.effect.Resource
+import japgolly.scalajs.react.*
+import cats.effect.std.Dispatcher
+import japgolly.scalajs.react.vdom.html_<^.*
+import crystal.react.hooks.*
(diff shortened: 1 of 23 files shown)


modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
```diff
@@ -0,0 +1,212 @@
...
+    _.flatMap:
+      case WSFrame.Text(text, _) => fs2.Stream(decode[ClientEvent](text))
+      case _                     => fs2.Stream.empty
+
+  // TODO IN CRYSTAL: It would be nice to have a .useEffectWhenDepReady instead of doing .void in dependencies!
+
+  private val component =
+    ScalaFnComponent
+      .withHooks[Unit]
+      .useResourceOnMount: // Build AppContext
```
cquiroz, line 127: This is a good idea, should we replicate in explore?

modules/web/client/vite.config.js
```diff
@@ -139,9 +139,10 @@ export default defineConfig(async ({ mode }) => {
       https: true,
       proxy: {
         '/api': {
-          target: 'http://localhost:7070',
+          target: 'http://127.0.0.1:7070',
```
cquiroz, line 142: maybe `local.lucuma.xyz` would work too?

cquiroz (approved):
This seems like a very. good idea

modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
rpiaggio: We could do in explore, yes

PR #813, opened 2023-10-05
Require SSO authentication to access the routes

cquiroz:
In this PR we require authentication for some of the routes, a token is passed with the POST requests and only valid users can access those routes


modules/model/shared/src/test/scala/observe/model/ModelSuite.scala
```diff
@@ -7,6 +7,7 @@ import cats.kernel.laws.discipline.*
 import lucuma.core.util.arb.ArbEnumerated.given
 import observe.model.GmosParameters.*
 import observe.model.ObserveModelArbitraries.given
+import observe.model.ObserveModelArbitraries.given_Arbitrary_Operator
```
rpiaggio, line 10: Isn't this import included in the previous one?

modules/web/server/src/main/resources/app.conf
```diff
@@ -46,7 +49,7 @@ web-server {
 # Configuration of the observe engine
 observe-engine {
     # host for the odb
-    odb = "ws://localhost:8082/ws"
+    odb = "wss://lucuma-postgres-odb-dev.herokuapp.com/ws"
```
rpiaggio, line 52: Can we also allow an override for this so we can use a local ODB? I guess an env variable will do.

modules/web/server/src/main/scala/observe/web/server/http4s/WebServerLauncher.scala
```diff
@@ -101,10 +105,30 @@ object WebServerLauncher extends IOApp with LogInitialization {
...
+        .initial(
+          serviceJwt = sso.serviceToken,
+          ssoRoot = sso.ssoUrl,
+          jwtReader = jwtReader[F](sso),
+          httpClient = httpClient
+        )
+        .map(_.map(_.user))
+    )
+
+  // SsoJwtReader(JwtDecoder.withPublicKey(sso.publicKey)).map { jwtReader =>
```
rpiaggio, line 126: Commented code

rpiaggio (approved):
LGTM

modules/model/shared/src/test/scala/observe/model/ModelSuite.scala
cquiroz: Ups

modules/web/server/src/main/resources/app.conf
cquiroz: I wonder how. with an env variable?

PR #818, opened 2023-10-10
Send the execution state on refresh

cquiroz:


modules/model/shared/src/main/scala/observe/model/client.scala
```diff
@@ -8,31 +8,46 @@ import cats.derived.*
...
+
+extension (v: SequencesQueue[SequenceView])
+  def sequencesState: SequencesExecutionState =
+    SequencesExecutionState(v.sessionQueue.map(o => (o.obsId, o.executionState)).toMap)
+
+extension (q: SequenceView)
+  def executionState: ExecutionState =
+    ExecutionState(q.status, q.runningStep.flatMap(_.id), None, Nil)
+
+case class SequencesExecutionState(sequences: Map[Observation.Id, ExecutionState])
```
rpiaggio, line 34: This can be an opaque type

rpiaggio (approved):
LGTM

PR #823, opened 2023-10-11
Restore the load route

cquiroz:


modules/web/server/src/main/scala/observe/web/server/http4s/vars.scala
```diff
@@ -4,9 +4,30 @@
...
+import observe.model.enums.Instrument
 
 object ClientIDVar {
   def unapply(str: String): Option[ClientId] =
     Either.catchNonFatal(ClientId(java.util.UUID.fromString(str))).toOption
 }
+
+trait EnumeratedVar[A: Enumerated] {
+  def unapply(str: String): Option[A] =
+    Enumerated[A].all.find(a => Enumerated[A].fromTag(str).exists(_ === a))
```
rpiaggio, line 20: Why not just use `Enumerated[A].fromTag` ?

cquiroz: uh

rpiaggio (approved):
LGTM

PR #830, opened 2023-10-13
Restore functionality to enable/disable subsystems

cquiroz:
These were commented on the engine and are now restored
After thinking a bit I think most routes should take a clientId to log which client does what action.
This required some changes on the client code.

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/newTypes.scala
+++ b/modules/model/shared/src/main/scala/observe/model/newTypes.scala
@@ -25,3 +25,10 @@ type Operator = Operator.Type
 
 object ImageFileId extends NewType[String]
 type ImageFileId = ImageFileId.Type
+
+object SubsystemEnabled extends NewType[Boolean] {
+  val Enabled  = SubsystemEnabled(true)
+  val Disabled = SubsystemEnabled(false)
+}
+
+type SubsystemEnabled = SubsystemEnabled.Type
--- a/modules/server_new/src/main/scala/observe/server/SeqTranslate.scala
+++ b/modules/server_new/src/main/scala/observe/server/SeqTranslate.scala
@@ -777,27 +777,27 @@ object SeqTranslate {
 //    private val gnirsDisabled: GnirsController[F]           = new GnirsControllerDisabled[F]
 
     def tcsSouth(overrides: SystemOverrides): TcsSouthController[F] =
-      if (overrides.isTcsEnabled) systems.tcsSouth
+      if (overrides.isTcsEnabled.value) systems.tcsSouth
       else tcsSouthDisabled
 
     def tcsNorth(overrides: SystemOverrides): TcsNorthController[F] =
-      if (overrides.isTcsEnabled) systems.tcsNorth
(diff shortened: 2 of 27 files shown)


modules/model/shared/src/main/scala/observe/model/newTypes.scala
```diff
@@ -25,3 +25,10 @@ type Operator = Operator.Type
 
 object ImageFileId extends NewType[String]
 type ImageFileId = ImageFileId.Type
+
+object SubsystemEnabled extends NewType[Boolean] {
```
rpiaggio, line 29: Do we need the boolean value? Otherwise, this can just be an `enum`.

modules/server_new/src/main/scala/observe/server/SeqTranslate.scala
```diff
@@ -777,39 +777,39 @@ object SeqTranslate {
...
       else tcsNorthDisabled
 
     def gems(overrides: SystemOverrides): GemsController[F] =
-      if (overrides.isTcsEnabled) systems.gems
+      if (overrides.isTcsEnabled.value) systems.gems
       else gemsDisabled
 
     def altair(overrides: SystemOverrides): AltairController[F] =
-      if (overrides.isTcsEnabled) systems.altair
+      if (overrides.isTcsEnabled.value) systems.altair
```
rpiaggio, line 792: Ah, I see here that the boolean value is used. We could check for ` === Enabled` instead.

Anyway, I don't see clear advantages or disadvantages of either method.

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
```diff
@@ -129,45 +129,56 @@ class ObserveCommandRoutes[F[_]: Async: Compression](
...
-    //     Ok(s"Set GCAL enable flag to '$gcalEnabled' for sequence $obsId")
-    //
-    // case POST -> Root / ObsId(obsId) / "instEnabled" / BooleanVar(instEnabled) as user =>
-    //   se.setInstrumentEnabled(inputQueue, obsId, user, instEnabled) *>
-    //     Ok(s"Set instrument enable flag to '$instEnabled' for sequence $obsId")
-    //
-    // case POST -> Root / ObsId(obsId) / "dhsEnabled" / BooleanVar(dhsEnabled) as user =>
-    //   se.setDhsEnabled(inputQueue, obsId, user, dhsEnabled) *>
-    //     Ok(s"Set DHS enable flag to '$dhsEnabled' for sequence $obsId")
+    case req @ POST -> Root / ObsIdVar(obsId) / ClientIDVar(clientId) / "tcsEnabled" /
```
rpiaggio, line 132: Should we consider passing the `clientId` in a header instead?

rpiaggio (approved):
LGTM

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
cquiroz: I don't know any pros/cons?

PR #831, opened 2023-10-13
Restore endpoint to set a breakpoint

cquiroz:


modules/engine/src/main/scala/observe/engine/Step.scala
```diff
@@ -24,16 +25,17 @@ final case class Step[F[_]](
...
 object Step {
 
-  final case class BreakpointMark(self: Boolean) extends AnyVal
-
-  final case class SkipMark(self: Boolean) extends AnyVal
-
-  final case class Skipped(self: Boolean) extends AnyVal
+  object BreakpointMark extends NewType[Boolean]
+  type BreakpointMark = BreakpointMark.Type
+  object SkipMark extends NewType[Boolean]
```
rpiaggio, line 30: I think we decided not to have skipped steps anymore. Maybe we can remove the logic? It can be another PR.

rpiaggio (approved):
LGTM

modules/engine/src/main/scala/observe/engine/Step.scala
cquiroz: Ok, I'll do that, there is some leftover code on the engine about them

PR #832, opened 2023-10-13
Instrument unification

rpiaggio:
Removes the `Resource.Instrument` sub sum type in favor of the `Instrument` enum from core.

`Resource` can be declared as an `enum` now.

Where necessary, `Resource` type is changed to `Resource | Instrument`.

Maybe we can made an opaque `System` type alias, but there's already a `System` type that encapsulates a `Resource | Instrument`, which may be converted into extension methods. We can evaluate this in a future PR.


cquiroz (approved):
LGTM

PR #833, opened 2023-10-15
Edit Observer. Abstract clientId in APIs. New SequenceApi. Invoke setBreakpoint on server.

rpiaggio:
`Observer` and `Operator` were changed to `Option[NonEmptyString]`.

Observer can be edited in `ConfigPanel`.

`clientId` is passed on construction to APIs instead of on every method.

New `SequenceApi` which for the moment only allows toggling breakpoints. The endpoint is now called when toggling a breakpoint in the UI.


modules/server_new/src/main/scala/observe/server/keywords/StandardHeader.scala
```diff
@@ -26,9 +26,8 @@ final case class StateKeywordsReader[F[_]: Applicative](
   observer:   Option[Observer]
 ) {
 
-  def observerName: F[String]       =
-    observer.map(_.value).filter(_.nonEmpty).getOrElse("observer").pure[F]
-  def operatorName: F[String]       = operator.map(_.value).filter(_.nonEmpty).getOrElse("ssa").pure[F]
+  def observerName: F[String]       = observer.map(_.value.value).getOrElse("observer").pure[F]
+  def operatorName: F[String]       = operator.map(_.value.value).getOrElse("ssa").pure[F]
```
cquiroz, line 30: Maybe `sso` or `operator`

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTables.scala
```diff
@@ -208,7 +209,15 @@ private sealed trait SequenceTablesBuilder[S: Eq, D: Eq]:
                 <.div(
                   <.div(
                     ObserveStyles.BreakpointHandle,
-                    stepId.map(^.onClick --> props.flipBreakpoint(_)).whenDefined
+                    stepId
```
cquiroz, line 212: Do we wait for the server to indicate the breakpont is enabled or is it optimistically set on the client?

cquiroz (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTables.scala
rpiaggio: For the moment, it is optimistically set. However, if the invocation results in an error, the UI will enter resync mode.

Anyway, maybe we should check for the server's response and (un)set the breakpoint again according the server's value. I will do that.

modules/server_new/src/main/scala/observe/server/keywords/StandardHeader.scala
rpiaggio: observer is initialized to sso in the UI. Should we set the operator too?

cquiroz: nope, we don't really know who the operator is

cquiroz: The observer should be set to the name of the logged in user

rpiaggio: Yup, that's done in the UI. Then it can be edited, but not left blank.

PR #836, opened 2023-10-17
Restore methos to set operator and observer

cquiroz:
operator and observer will be included on the events stream too


rpiaggio (approved):
LGTM

PR #838, opened 2023-10-18
Show observation header, keep obs details and sequence loaded until a new one is selected

rpiaggio:
Moves the loaded sequence to the `RootModel`, as well as providing global hooks to update it, as well as the observation details, when a new observation is selected anywhere, freeing the component logic from this task, and avoiding a requery in case of navigating to another tab and back.

I needed to create a model class `ObsSummary` to hold observation details which needed to be referenced from clue queries. Therefore, I had to place it in a new project.

The observation header displays basic information for the moment and it's lacking more stylized... well, styling.

### DIFF
--- a/modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
+++ b/modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
@@ -1,0 +1,83 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package observe.ui.model
+
+import cats.Eq
+import cats.Order.given
+import cats.derived.*
+import cats.syntax.all.*
+import io.circe.Decoder
+import io.circe.generic.semiauto.*
+import lucuma.core.model.ConstraintSet
+import lucuma.core.model.ObsAttachment
+// import lucuma.core.model.Observation
+import lucuma.core.model.PosAngleConstraint
+import lucuma.core.model.TimingWindow
+import lucuma.core.util.Timestamp
+import lucuma.schemas.decoders.given
+import lucuma.schemas.model.BasicConfiguration
+import lucuma.schemas.model.ObservingMode
+import monocle.Focus
+import org.typelevel.cats.time.*
+
+import java.time.Instant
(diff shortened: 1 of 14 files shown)


cquiroz:
Is that like `ObsSummary` in explore?

modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
```diff
@@ -0,0 +1,76 @@
...
+import lucuma.schemas.decoders.given
+import lucuma.schemas.model.BasicConfiguration
+import lucuma.schemas.model.ObservingMode
+import monocle.Focus
+import org.typelevel.cats.time.*
+
+import java.time.Instant
+import scala.collection.immutable.SortedSet
+
+case class ObsSummary(
```
cquiroz, line 26: I wonder if this can be shared with explore, we have a very similar class over there

modules/web/client/src/main/scala/observe/ui/components/queue/SessionQueue.scala
```diff
@@ -219,8 +221,8 @@ object SessionQueue:
             tableMod = ObserveStyles.ObserveTable |+| ObserveStyles.SessionTable,
             rowMod = row =>
               TagMod(
-                rowClass( /*row.index.toInt,*/ row.original, props.selectedObsId.get),
-                ^.onClick --> props.selectedObsId.set(row.original.obsId.some)
+                rowClass( /*row.index.toInt,*/ row.original, props.selectedObsId),
+                ^.onClick --> props.setSelectedObsId(row.original.obsId)
```
cquiroz, line 225: load on click then? what if it is running?

cquiroz (approved):
LGTM

modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
rpiaggio: That was my initial hunch but after checking, it turns out we request way less information here.

modules/web/client/src/main/scala/observe/ui/components/queue/SessionQueue.scala
rpiaggio: I will switch to a load button in a future PR.

PR #840, opened 2023-10-19
Support setting multiple breakpoints and set breakpoints on load

cquiroz:
breakpoints are read from the odb on load and it is possible to set multiple breakpoints at once in a separate `breakpoints` route

### DIFF
--- a/modules/engine/src/main/scala/observe/engine/Event.scala
+++ b/modules/engine/src/main/scala/observe/engine/Event.scala
@@ -6,6 +6,7 @@ package observe.engine
 import cats.effect.Sync
 import cats.syntax.all.*
 import fs2.Stream
+import lucuma.core.enums.Breakpoint
 import lucuma.core.model.User
 import observe.engine.SystemEvent.Null
 import observe.model.ClientId
@@ -33,11 +34,11 @@ object Event {
   def cancelPause[F[_], S, U](id: Observation.Id, user: User): Event[F, S, U]               =
     EventUser[F, S, U](CancelPause(id, user.some))
   def breakpoint[F[_], S, U](
-    id:   Observation.Id,
-    user: User,
-    step: StepId,
-    v:    Boolean
-  ): Event[F, S, U] = EventUser[F, S, U](Breakpoint(id, user.some, step, v))
+    id:    Observation.Id,
+    user:  User,
+    steps: List[StepId],
+    v:     Breakpoint
+  ): Event[F, S, U] = EventUser[F, S, U](Breakpoints(id, user.some, steps, v))
   def skip[F[_], S, U](
     id:   Observation.Id,
     user: User,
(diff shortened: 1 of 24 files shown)


modules/engine/src/main/scala/observe/engine/Event.scala
```diff
@@ -33,11 +34,11 @@ object Event {
   def cancelPause[F[_], S, U](id: Observation.Id, user: User): Event[F, S, U]               =
     EventUser[F, S, U](CancelPause(id, user.some))
   def breakpoint[F[_], S, U](
```
rpiaggio, line 36: `def breakpoints` ? (with s)

cquiroz: yes, it takes a list

modules/model/shared/src/main/scala/observe/model/Step.scala
```diff
@@ -32,6 +33,10 @@ object Step {
   extension [A](l: Lens[A, Boolean]) {
     def negate: A => A = l.modify(!_)
```
rpiaggio, line 34: Do we still need `negate` for other things?

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
```diff
@@ -71,11 +73,20 @@ class ObserveCommandRoutes[F[_]: Async: Compression](
...
       ssoClient.require(req) { user =>
-        oe.setBreakpoint(obsId, user, obs, stepId, bp) *>
+        oe.setBreakpoints(obsId, user, obs, List(stepId), bp) *>
           NoContent()
       }
+
+    case req @ POST -> Root / ObsIdVar(obsId) / ClientIDVar(clientId) /
+        "breakpoints" / ObserverVar(obs) / BreakpointVar(bp) =>
+      ssoClient.require(req) { user =>
+        req.decode[List[Step.Id]] { steps =>
```
rpiaggio, line 85: How is this decoded? Just a comma-separated list of step ids?

rpiaggio (approved):
LGTM

modules/model/shared/src/main/scala/observe/model/Step.scala
cquiroz: it is being called bet we'll probablyy remove it when we take the skipmark off

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
cquiroz: Yes like `["s-f5efec44-32f4-3b6b-8e9a-39dcf55ed31c", "s-800d515f-835d-3522-aca3-1f21b2865d84"]`

PR #841, opened 2023-10-20
Get all state from server, start sequence working

rpiaggio:
In this PR initial state is now completely obtained from the server (instead of using a dummy one).

Logic has been added to synchronize connections to both ODB and Observe server.

"Start" command is now working and all clients connected are being synced.

PR #843, opened 2023-10-20
Pass initial breakpoints from ODB to server when loading a sequence

rpiaggio:
This PR subsumes #841. It will be easier to review after merging that other PR.

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/services/ObservationSyncer.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/services/ObservationSyncer.scala
@@ -8,6 +8,10 @@ import cats.syntax.all.*
 import crystal.react.*
 import japgolly.scalajs.react.*
 import japgolly.scalajs.react.vdom.html_<^.*
+import lucuma.core.enums.Breakpoint
+import lucuma.core.model.sequence.ExecutionSequence
+import lucuma.core.model.sequence.InstrumentExecutionConfig
+import lucuma.core.model.sequence.Step
 import lucuma.react.common.ReactFnProps
 import lucuma.schemas.odb.SequenceSQL
 import lucuma.ui.reusability.given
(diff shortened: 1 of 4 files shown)


modules/web/client/src/main/scala/observe/ui/components/services/ObservationSyncer.scala
```diff
@@ -64,8 +68,27 @@ object ObservationSyncer:
...
+                    config match
+                      case InstrumentExecutionConfig.GmosNorth(executionConfig) =>
+                        getBreakPoints(executionConfig.acquisition) ++
+                          getBreakPoints(executionConfig.science)
+                      case InstrumentExecutionConfig.GmosSouth(executionConfig) =>
+                        getBreakPoints(executionConfig.acquisition) ++
+                          getBreakPoints(executionConfig.science)
+
+                  sequenceApi.loadObservation(obsId, config.instrument) >>
+                    sequenceApi.setBreakpoints(obsId, initialBreakpoints.toList, Breakpoint.Enabled)
```
cquiroz, line 91: So we are submitting breakpoints from the odb on load? I think we are doing thay already on the odb

rpiaggio: Yes. This sets the initial breakpoints for the portion of the sequence that the UI reads (max 100 atoms). Does the server read all that?

cquiroz: We only read one atom, thus this call will set the breakpoint for the first atom and ignore the rest

cquiroz (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/components/services/ObservationSyncer.scala
rpiaggio: Can we keep the set of breakpoints in the server even if it hasn't read the atoms yet? Otherwise we would disable setting breakpoints beyond the current atom in the UI.

PR #846, opened 2023-10-23
Allow selection and unselection of rows in UI when nothing is running

rpiaggio:
When a step is selected, step controls are shown. This functionality is disabled when something is running.

Note that there's still of unported commented code around the modified code.


cquiroz (approved):
LGTM

PR #849, opened 2023-10-24
Execute subsystems from UI

rpiaggio:
For the moment, I hard coded TCS, GCal and Gmos. Let's switch when resources come in the state from the server.


modules/model/shared/src/main/scala/observe/model/package.scala
```diff
@@ -66,3 +66,12 @@ given KeyEncoder[Resource | Instrument] = KeyEncoder.instance:
   case i: Instrument => i.tag
 given KeyDecoder[Resource | Instrument] = KeyDecoder.instance: tag =>
   Enumerated[Resource].fromTag(tag).orElse(Enumerated[Instrument].fromTag(tag))
+
+given Enumerated[Resource | Instrument] = Enumerated
```
cquiroz, line 70: nice

cquiroz (approved):
LGTM

modules/model/shared/src/main/scala/observe/model/package.scala
rpiaggio: It was your idea :)

PR #855, opened 2023-10-26
Include step resources being executed as part of the state

cquiroz:
We want to send the state of each subsystem when they are triggered individually.
This requires some changes to the engine and server to carry the state of each subsystem all the way to the client
Note that wee are  only sending the state for the current atom


modules/web/client/src/main/scala/observe/ui/components/services/ServerEventHandler.scala
```diff
@@ -71,14 +83,14 @@ trait ServerEventHandler:
...
                   executionStatus.copy(
-                    configStatus =
-                      List[Resource | Instrument](Resource.TCS, Resource.Gcal, Instrument.GmosNorth)
-                        .foldLeft(executionStatus.configStatus)((cs, ss) =>
-                          cs.updatedWith(ss)(
-                            _.orElse(old.get(obsId).flatMap(_.configStatus.get(ss)))
-                              .orElse(ActionStatus.Pending.some)
-                          )
-                        )
+                    stepResources = Nil // TODO Fixme
```
cquiroz, line 86: This is an exercise for @rpiaggio 😄

jluhrs (commented):
I don't really understand why this change is needed.

modules/web/client/src/main/scala/observe/ui/components/services/ServerEventHandler.scala
rpiaggio: Lol, ok

modules/engine/src/main/scala/observe/engine/Engine.scala
```diff
@@ -96,14 +95,10 @@ class Engine[F[_]: MonadThrow: Logger, S, U] private (
...
-               case ActionType.Configure(sys) => sys.some
-               case _                         => none
-             }
-    } yield (act.gen, sys)
-
-    x.map { case (p, s) =>
-      modifyS(c.sid)(u => u.startSingle(c.actCoords, Response.Configured(s))) *>
+    } yield act.gen
+
+    x.map { case p =>
```
jluhrs, line 100: case p ?

jluhrs (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/components/Home.scala
```diff
@@ -158,10 +160,11 @@ object Home:
                             else
                               // TODO We actually have to remember what step this belonged to before it was unselected.
                               (stepId.some,
-                               executionState.map(
-                                 ExecutionState.configStatus
-                                   .modify(_.map((k, _) => k -> ActionStatus.Pending))
-                               )
+                               executionState.map(identity)
+                               // TODO Fixme
```
cquiroz, line 164: Another exercise for @rpiaggio

PR #862, opened 2023-11-07
Better error display, adapt to latest server changes reporting subsystem state

rpiaggio:


modules/model/shared/src/main/scala/observe/model/ExecutionState.scala
```diff
@@ -28,7 +26,7 @@ case class ExecutionState(
   observer:        Option[Observer],
   runningStepId:   Option[Step.Id],
   nsState:         Option[NsRunningState],
-  stepResources:   List[(Step.Id, List[(Resource | Instrument, ActionStatus)])],
+  stepResources:   Map[Step.Id, Map[Resource | Instrument, ActionStatus]],
```
rpiaggio, line 29: I found it easier to deal with this if we model as a `Map` (allows for example `index`/`at` optics).

modules/model/shared/src/main/scala/observe/model/ExecutionState.scala
```diff
@@ -42,25 +40,12 @@ case class ExecutionState(
...
+  val nsState: Lens[ExecutionState, Option[NsRunningState]]                                       = Focus[ExecutionState](_.nsState)
+  val stepResources: Lens[ExecutionState, Map[Step.Id, Map[Resource | Instrument, ActionStatus]]] =
     Focus[ExecutionState](_.stepResources)
-  val systemOverrides: Lens[ExecutionState, SystemOverrides] =
+  val systemOverrides: Lens[ExecutionState, SystemOverrides]                                      =
     Focus[ExecutionState](_.systemOverrides)
-  val breakpoints: Lens[ExecutionState, Set[Step.Id]]        = Focus[ExecutionState](_.breakpoints)
-  val observer: Lens[ExecutionState, Option[Observer]]       = Focus[ExecutionState](_.observer)
-
-  def stepResourcesT(
```
rpiaggio, line 56: By using `index`/`at` this was left unused so I removed it.

modules/model/shared/src/main/scala/observe/model/StepState.scala
```diff
@@ -16,35 +16,35 @@ enum StepState derives Eq:
   case Running             extends StepState
   case Paused              extends StepState
 
-  def canSetBreakpoint: Boolean = this match
+  lazy val canSetBreakpoint: Boolean = this match
```
rpiaggio, line 19: I was having a strange lazy initialization error regarding the derived `Eq`. Thought changing to `lazy val`s would fix it, but it didn't (`sbt clean` did). I decided to leave these as `lazy val`s anwyay.

modules/web/client/src/main/webapp/styles/observe.scss
```diff
@@ -166,58 +166,71 @@ html {
   }
 }
 
-// .ObserveStyles-rowPositive {
-//   background-color: @positiveBackgroundColor;
-//   color: @positiveTextColor;
-//   box-shadow: @positiveBoxShadow;
-// }
+.pl-react-table.p-datatable.p-datatable-hoverable-rows .p-datatable-tbody {
+  & > tr {
```
rpiaggio, line 170: Commented styles are still pending revision from seqexec migration.

modules/model/shared/src/main/scala/observe/model/ExecutionState.scala
cquiroz: 👍

modules/model/shared/src/main/scala/observe/model/StepState.scala
cquiroz: I wonder if these should be `inline`

cquiroz (approved):
LGTM

PR #871, opened 2023-11-14
Pause events

cquiroz:
Before we were forwarding some of the model update events but this changes will send them all, e.g. when a sequence is paused or removed
Stop/pause/etc.. aren't step specific so I removed `stepId` from the routes

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/events.scala
+++ b/modules/model/shared/src/main/scala/observe/model/events.scala
@@ -138,28 +138,8 @@ case class AlignAndCalibEvent(step: Int) extends ObserveEvent derives Eq
 
 extension (e: ObserveEvent)
   def toClientEvent: Option[(Option[ClientId], ClientEvent)] = (e match {
-    case ConditionsUpdated(v)                                   =>
-      ClientEvent.ObserveState(v.sequencesState, v.conditions, v.operator).some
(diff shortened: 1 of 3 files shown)


modules/model/shared/src/main/scala/observe/model/events.scala
```diff
@@ -138,28 +138,8 @@ case class AlignAndCalibEvent(step: Int) extends ObserveEvent derives Eq
...
-      ClientEvent.ObserveState(v.sequencesState, v.conditions, v.operator).some
-    case FileIdStepExecuted(_, v)                               =>
-      ClientEvent.ObserveState(v.sequencesState, v.conditions, v.operator).some
-    case SequenceStart(_, _, v)                                 =>
-      ClientEvent.ObserveState(v.sequencesState, v.conditions, v.operator).some
-    case SequenceCompleted(v)                                   =>
-      ClientEvent.ObserveState(v.sequencesState, v.conditions, v.operator).some
-    case OperatorUpdated(v)                                     =>
-      ClientEvent.ObserveState(v.sequencesState, v.conditions, v.operator).some
+    case a: ObserveModelUpdate                                  =>
```
rpiaggio, line 141: nice

rpiaggio (approved):
LGTM

PR #886, opened 2023-11-27
Show progress in the client

rpiaggio:


modules/model/shared/src/main/scala/observe/model/ObservationProgress.scala
```diff
@@ -4,42 +4,25 @@
...
-  case NodAndShuffle(
-    obsId:     Observation.Id,
-    stepId:    Step.Id,
-    total:     TimeSpan,
-    remaining: TimeSpan,
-    stage:     ObserveStage,
-    sub:       NsSubexposure
-  ) extends ObservationProgress(true)
+case class ObservationProgress(obsId: Observation.Id, stepProgress: StepProgress):
+  export stepProgress.*
```
cquiroz, line 13: What does this do?

rpiaggio: Allows direct access to `StepProgress` members. I thought this might be convenient since I split `ObservationProgress` into 2 classes.

cquiroz (approved):
LGTM

PR #887, opened 2023-11-28
Embed JRE with deployed package

rpiaggio:
I simplified the logic since we are only deploying for linux-x64. If we need to support multiple environments, we can add the logic back.

For the moment, a single artifact is still built for all environments, which should be enough for upcoming tests. Differentiation of site and test/prod upcoming in future, not urgent, PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,9 +34,9 @@ ThisBuild / resolvers += "Gemini Repository".at(
   "https://github.com/gemini-hlsw/maven-repo/raw/master/releases"
 )
 
-// This key is used to find the JRE dir. It could/should be overridden on a user basis
-// Add e.g. a `jres.sbt` file with your particular configuration
-ThisBuild / ocsJreDir := Path.userHome / ".jres11"
+// This key is used to find the JRE dir that will be bundled with packaged deployments.
+// It could/should be overridden on a user basis. E.g. add a `jres.sbt` file with your particular configuration.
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -34,9 +34,9 @@ ThisBuild / resolvers += "Gemini Repository".at(
   "https://github.com/gemini-hlsw/maven-repo/raw/master/releases"
 )
 
-// This key is used to find the JRE dir. It could/should be overridden on a user basis
-// Add e.g. a `jres.sbt` file with your particular configuration
-ThisBuild / ocsJreDir := Path.userHome / ".jres11"
+// This key is used to find the JRE dir that will be bundled with packaged deployments.
+// It could/should be overridden on a user basis. E.g. add a `jres.sbt` file with your particular configuration.
+ThisBuild / packagedJreDir := Path.userHome / ".jres" / "linux" / "jdk-11.0.21+9-jre"
```
cquiroz, line 39: I wonder if we should go 17

cquiroz (approved):
LGTM

build.sbt
cquiroz: In the past we had a link like `.jres/linux/JDK` wheere you locally select the JVM changing the link. WDYT?

cquiroz: That would be fine. Could you add a note in the README?

PR #899, opened 2023-12-06
Some fixes for deployment

cquiroz:


app/observe-server/src/main/resources/conf/app.conf
rpiaggio, line 1: This file was supposed to have fallback defaults for all servers. Do we want to delete it?

build.sbt
```diff
@@ -127,7 +127,7 @@ lazy val deployedAppMappings = Seq(
 
   // The only thing we include from the base deployment app is app.conf. We remove the "conf" path.
   Compile / packageBin / mappings ~= {
-    _.filter(_._1.getName.endsWith(".conf")).map(mapping => mapping._1 -> mapping._1.getName)
+    _.filter(_._1.getName.endsWith(".conf.json")).map(mapping => mapping._1 -> mapping._1.getName)
```
rpiaggio, line 130: Why this?

app/observe-server/src/main/resources/conf/app.conf
cquiroz: Actually the fallback is on the classpath inside a jar

build.sbt
cquiroz: Because it wasn't including `environments.conf.json` which is what we need

app/observe-server/src/main/resources/conf/app.conf
rpiaggio: Exactly, I think this is the app.conf that ends up in the jar.

build.sbt
rpiaggio: This is a common mapping for all 4 deployments. Do we want a common `environments.conf.json` for all 4 deployments?

PR #907, opened 2023-12-12
Stop+abort sequence, preload observation summaries into state

rpiaggio:
- Enable stop and abort buttons at the step level. Sometimes these buttons are enabled when they shouldn't. This will be addressed in future PR, I'm still working on migrating the logic from seqexec.
- Preload observation summaries into state. The idea here is to show the observation information when an observation is selected in the table, without the need of loading it into observe. ~That part of the UI is still pending for the next PR.~ Information is shown now, and the sequence can be loaded from the header too.

An issue we have here is that, after a sequence is loaded, preview functionality is lost since now the observation header is locked into the loaded sequence. We probably want to display observation details from the list somewhere else. A tooltip? A side panel? I'm open to suggestions.

This probably should have been 2 separate PRs, sorry for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ stopObserveAll
 
 # How to package for deployment
 
-If you haven't already, create a symlink `jre` that points to the JRE you want bundled with the deployment.
+If you haven't already, create a symlink `jre` in the project directory that points to the Linux JRE you want bundled with the deployment.
 
 To deploy, run in `sbt`:
 
--- a/modules/model/shared/src/main/scala/observe/model/operations.scala
+++ b/modules/model/shared/src/main/scala/observe/model/operations.scala
@@ -47,7 +47,6 @@ object operations:
       isObservePaused: Boolean,
       isMultiLevel:    Boolean
     ): List[Operations] =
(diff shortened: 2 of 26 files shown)


cquiroz:
Could you elaborate?

> An issue we have here is that, after a sequence is loaded, preview functionality is lost since now the observation header is locked into the loaded sequence. We probably want to display observation details from the list somewhere else. A tooltip? A side panel? I'm open to suggestions.

cquiroz (approved):
LGTM

rpiaggio:
image


When you select a row in the table above, the observation details are shown below. But if a sequence is loaded, the details are locked to its observation, and you cannot see the details of other rows. Maybe we should show the details somewhere else? Put all the information in the table? That table is not in the inception doc though, one day it will substituted with a chart with the scheduler's plan for the night.

> Could you elaborate?
> 
> > An issue we have here is that, after a sequence is loaded, preview functionality is lost since now the observation header is locked into the loaded sequence. We probably want to display observation details from the list somewhere else. A tooltip? A side panel? I'm open to suggestions.

PR #908, opened 2023-12-13
Reconnect to server and resync

rpiaggio:
In case the WS connection to the server is broken, a reconnect procedure is started, and state is resynced upon reconnection.

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -25,7 +25,7 @@ object Settings {
     val scalaXml          = "1.2.0"
     val catsTime          = "0.5.1"
 
-    val http4sVersion              = "0.23.24"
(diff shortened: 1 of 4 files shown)


project/Settings.scala
```diff
@@ -25,7 +25,7 @@ object Settings {
     val scalaXml          = "1.2.0"
     val catsTime          = "0.5.1"
 
-    val http4sVersion              = "0.23.24"
+    val http4sVersion              = "0.23.24-25-e71f1b1-SNAPSHOT"
```
cquiroz, line 28: Why do we need a snapshot?

cquiroz (approved):
Thanks, LGTM

project/Settings.scala
rpiaggio: The `Reconnect` middleware for WS is not released yet.

PR #912, opened 2023-12-22
Keep track of UI requests

rpiaggio:
This PR keeps track of UI requests and enables/disables buttons accordingly.

For the moment, all requests for all observations are reset as soon as any state update is received from the server. We may want a finer grained behavior here, but in order to achieve that we have to either compare the new state with our current one in order to know what has changed in the server, or have finer grained events than full state updates. On the other hand, the current behavior may be enough.

The class `ObservationRequests` kind of replaces the previous `TabOperations`. `TabOperations` used to have specialized types for each request, which was nice, but I opted for a common `enum` here in order to be able to have common logic over them (see `SequenceApiImpl`).


cquiroz (approved):
LGTM

PR #927, opened 2024-01-09
Remove ldap code

rpiaggio:


cquiroz:
I guess we won't be using LDAP though it maybe the preferred for IT operations
Let's remove it and if needed it we can resurrect it

cquiroz (commented):
LGTM Thanks

PR #928, opened 2024-01-10
Docker deployment

rpiaggio:
This PR substitutes the testing/production deployment method to Docker images.

Furthermore, it adds the possibility of building staging images with a server in simulation mode.

Stating images are pushed to the Heroku app.

The base deployment app is now the staging app. Testing and production deployments just override a minimal set of properties. For the moment, overrides are only defined for `gn_test`.

For clarity, the `app` family of projects and folder were renamed to `deploy`.

In a future PR, I will attempt to auto build and push the Docker image from CI when a PR is merged onto `main`.


README.md
```diff
@@ -70,15 +70,70 @@ and to stop all the processes you can do
...
+Make sure you have both `docker` and `heroku` CLIs installed and working.
+
+If you haven't already, run
+
+```
+heroku login
+heroku container:login
+```
+
+This will give your system access to Heroku's Docker registry.
```
cquiroz, line 88: What do we need to use heroku's registry from inside the gemini network?

README.md
```diff
@@ -70,15 +70,70 @@ and to stop all the processes you can do
...
+
+To deploy to Heroku, run in `sbt`:
+
+```
+deploy_observe_server/docker:publish
```
cquiroz, line 93: Is it possible to deploy locally?

README.md
rpiaggio: We will probably use Dockerhub for testing/production deployment, not Heroku's registry. I think we just need access to `hub.docker.com:443`, but I'll double check.

README.md
rpiaggio: Yes! You can `docker:publishLocal`. To run the container you just need to pass `SSO_SERVICE_JWT` (for staging SSO).

In fact, `publish` does a `publishLocal` folowed by a `push`.

deploy/observe-server-gn-test/src/main/resources/app/environments.conf.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "hostName": "observe.hi.gemini.edu",
+    "hostName": "*",
```
cquiroz, line 3: Can you explain this change?

modules/model/jvm/src/main/scala/observe/model/config/Mode.scala
```diff
@@ -6,15 +6,9 @@ package observe.model.config
...
-
-object Mode {
-  case object Production  extends Mode("Production")
-  case object Development extends Mode("Development")
-
-  given Enumerated[Mode] =
-    Enumerated.from(Production, Development).withTag(_.tag)
-
-}
+enum Mode(val tag: String) derives Enumerated:
```
cquiroz, line 11: Should this live in `lucuma-core`?

rpiaggio: Actually, we have a similar one in lucuma-ui. Let's unify.

deploy/observe-server-gn-test/src/main/resources/app/environments.conf.json
rpiaggio: This is legacy logic from explore, where we use the host as a discriminator to know what environment we are in. Here we just have a different file for each environment. 

In any case, these files are going away soon since we plan to pass these parameters from the server to the client.

modules/model/jvm/src/main/scala/observe/model/config/Mode.scala
rpiaggio: I'll unify in the next PR with the configuration rework.

PR #929, opened 2024-01-10
Reduced duplicated step types

cquiroz:
We have three `Step` classes present in observe which confuses me all the time.
I propose we rename them
`Step` from core remains, while `Step` in engine becomes `EngineStep` and `Step` in the server becomes `ObserveStep`


modules/model/shared/src/main/scala/observe/model/ObserveStep.scala
```diff
@@ -19,8 +19,8 @@ import monocle.syntax.all.*
 import observe.model.dhs.*
 import observe.model.enums.*
 
-sealed trait Step extends Product with Serializable {
-  def id: StepId
+sealed trait ObserveStep extends Product with Serializable {
+  def id: ObserveStep.Id
```
rpiaggio, line 23: Why not just use `Step.Id`?

modules/model/shared/src/main/scala/observe/model/ObserveStep.scala
```diff
@@ -19,8 +19,8 @@ import monocle.syntax.all.*
 import observe.model.dhs.*
 import observe.model.enums.*
 
-sealed trait Step extends Product with Serializable {
-  def id: StepId
+sealed trait ObserveStep extends Product with Serializable {
```
rpiaggio, line 22: Is it possible to just use core's `Step` instead of `ObserveStep`? Or at least have `ObserveStep` wrap a `Step` and get `id`, `instConfig` and `stepConfig` from the wrapped `Step`?

rpiaggio (approved):
LGTM, with a couple of comments

modules/model/shared/src/main/scala/observe/model/ObserveStep.scala
cquiroz: That seems a bug, thanks for noticingit

modules/model/shared/src/main/scala/observe/model/ObserveStep.scala
cquiroz: They have very different structures as one describes the step while `ObserveStep` contains state of the current execution

rpiaggio: I see both share the `DynamicConfig` and `StepConfig` (and `id`).

Would it be possible for `ObserveStep` to wrap a (core) `Step`?

I'm in the process of making the server share the current atom configuration with the UI but I find that the server has `ObserveStep`s, not (core) `Step`s. The UI has all the logic in place (which is shared with explore) to display (core) `Step`s. If the server provided (core) `Step`s instead of `ObserveStep`s, it would be much easier in the UI.

jluhrs (approved):
LGTM

modules/model/shared/src/main/scala/observe/model/ObserveStep.scala
rpiaggio: Upon closer examination, it might not be that hard to create an adaptation layer for `ObserveStep` for the UI. I'll attempt this.

PR #931, opened 2024-01-11
Unify client and server configuration

rpiaggio:
Instead of configuring clients via `environments.conf.json`, the configuration is now shared by the server in the initialization message (which was renamed from `Environment` to `ClientConfig`).

This required a rework of the UI initialization, since now we need to connect to the server and then to the ODB, which was done the other way round before.


cquiroz (approved):
LGTM

PR #935, opened 2024-01-15
Share current steps with UI

rpiaggio:
Includes refactor of `ObserveStep` ADT into an `enum`.

This is just the shared model and server side. Client side will come in an imminent PR.


cquiroz (approved):
LGTM

PR #944, opened 2024-01-17
Auto deploy staging app to Heroku upon merge

rpiaggio:


build.sbt
```diff
@@ -8,17 +8,95 @@ import org.scalajs.linker.interface.ModuleSplitStyle
...
+    WorkflowStep.Checkout ::
+      WorkflowStep.SetupJava(githubWorkflowJavaVersions.value.toList.take(1)) :::
+      setupNodeNpmInstall :::
+      herokuLogin ::
+      sbtDeploy ::
+      herokuRelease ::
+      Nil,
+    scalas = List(scalaVersion.value),
+    javas = githubWorkflowJavaVersions.value.toList.take(1),
+    cond = Some(allConds(anyConds(mainCond, prCond), geminiRepoCond, isMergedCond))
```
armanbilge, line 98: I think this is all we need. Since we only want to publish commits to the main branch, is that right?

```suggestion
    cond = Some(allConds(mainCond, geminiRepoCond))
```

rpiaggio: Is `mainCond` true when it runs on the main branch, or when the PR wants to merge to the main branch too?

rpiaggio: Let's try it.

armanbilge: `mainCond` is true only when the commit is directly pushed to the `main` branch (i.e. in a merge commit). When a PR wants to merge into main, then all its commits must exist on a non-`main` branch, and therefore fail this condition.

armanbilge (approved):

PR #948, opened 2024-01-24
Keep recorded ids inside ODB proxy

rpiaggio:
This is a rework of the `OdbProxy` so that recorded ids are kept internally with no need to put them in the engine state.

The existing logic to keep some of the recorded ids in the state is still there, I will attempt to remove it in the next PR. That's why a couple of the `OdbProxy` methods still return ids. Please use `void` to call them for the moment.

Invocations to the methods of the `OdbProxy` can now be added from the engine.

(Bonus: fatal warnings)


modules/server_new/src/main/scala/observe/server/odb/OdbProxy.scala
```diff
@@ -0,0 +1,489 @@
...
+
+    override def sequenceStart(
+      obsId:        Observation.Id,
+      instrument:   Instrument,
+      sequenceType: SequenceType,
+      stepCount:    NonNegShort,
+      staticCfg:    StaticConfig
+    ): F[(VisitId, RecordedAtomId)] =
+      for {
+        // TODO Check that there are no current visits or atoms??
```
cquiroz, line 286: Remember we could be running multiple instruments in parallel

cquiroz (approved):
LGTM

modules/server_new/src/main/scala/observe/server/odb/OdbProxy.scala
rpiaggio: Ok. But in that case we use the same visit, atom and step ids? Or do any of them change?

cquiroz: I'd say all change, they are per sequence which will be different for each instrument

rpiaggio: Ok, but they have the same `Observation.Id`? Should we change the key to `(Observation.Id, Instrument)`?

cquiroz: They could be different observations

PR #950, opened 2024-01-24
Remove recorded ids from state

rpiaggio:
Subsumes #948


modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -402,45 +402,42 @@ object ObserveEngine {
     // Produce a Handle that will send a SequenceStart notification to the ODB, and produces the (sequenceId, stepId)
     // if there is a valid sequence with a valid current step.
     private def sequenceStart(
-      id: Observation.Id
+      obsId: Observation.Id
     ): HandlerType[F, Option[(Observation.Id, Step.Id)]] =
```
rpiaggio, line 406: I think we can simplify the signature here, since it's not triggering any further events anymore.

jluhrs (approved):
LGTM

PR #955, opened 2024-01-30
ODB Proxy: Track dataset ids, rework id tracking

rpiaggio:
- Add missing functionality of tracking dataset ids in the ODB Proxy.
- Id tracking model is reworked to better reflect reality: now you can't have i.e. a recorded step id without having an atom id.
- Check that visit, step and dataset ids are not already set when attempting to set them.
- Id tracking Map is now an opaque type.
- Tracking methods have been extracted to a trait in another file to avoid clutter.
- Multiple concurrent dataset ids are now supported.
- Unused parameters have been removed.
- Methods have also been sorted to reflect the flow.


jluhrs (approved):
LGTM

PR #974, opened 2024-02-13
Emit ODB events while running an observation

jluhrs:


modules/server_new/src/main/scala/observe/server/SequenceData.scala
```diff
@@ -11,14 +11,36 @@ import observe.model.SystemOverrides
...
-)
+  observer:       Option[Observer],
+  overrides:      SystemOverrides,
+  seqGen:         SequenceGen[F],
+  seq:            Sequence.State[F],
+  pendingObsCmd:  Option[PendingObserveCmd],
+  visitStartDone: Boolean,
+  atomStartDone:  Boolean
+) {
+  def completeVisitStart: SequenceData[F] = this.copy(visitStartDone = true)
```
rpiaggio, line 22: Since these methods are not actually doing any action (completing things), may I suggest renaming to `withVisitStartDone` and `withAtomStartDone` ?

rpiaggio (approved):
LGTM. Nitpicking comment.

PR #979, opened 2024-02-16
Unbump core, fileId simulated logic

rpiaggio:
I had to dial back `lucuma-core` since the latest version renames a `DatasetStage` instance, producing incompatibilities.

Also, I make sure the DHS simulator produces filenames that do not conflict (implements @jluhrs 's idea).

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/keywords/DhsClientHttp.scala
+++ b/modules/server_new/src/main/scala/observe/server/keywords/DhsClientHttp.scala
@@ -5,6 +5,7 @@ package observe.server.keywords
 
 import cats.FlatMap
 import cats.data.EitherT
+import cats.effect.Clock
 import cats.effect.Ref
 import cats.effect.Sync
 import cats.effect.Temporal
@@ -30,7 +31,10 @@ import org.http4s.client.middleware.RetryPolicy
 import org.http4s.dsl.io.*
 import org.typelevel.log4cats.Logger
 
+import java.time.Instant
 import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import scala.concurrent.duration.*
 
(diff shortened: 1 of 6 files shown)


jluhrs:
Just be aware that there will be still a conflict if two fileIds are requested less than 10 seconds apart.

modules/server_new/src/main/scala/observe/server/keywords/DhsClientHttp.scala
```diff
@@ -238,9 +243,15 @@ private class DhsClientSim[F[_]: FlatMap: Logger](date: LocalDate, counter: Ref[
...
 object DhsClientSim {
   def apply[F[_]: Sync: Logger]: F[DhsClient[F]] =
-    (Sync[F].delay(LocalDate.now), Ref.of[F, Int](0)).mapN(new DhsClientSim[F](_, _))
+    Clock[F].monotonic
+      .map(d => Instant.EPOCH.plusNanos(d.toNanos))
+      .map(LocalDateTime.ofInstant(_, ZoneId.systemDefault))
+      .flatMap(apply)
 
-  def apply[F[_]: Sync: Logger](date: LocalDate): F[DhsClient[F]] =
-    Ref.of[F, Int](0).map(new DhsClientSim[F](date, _))
```
jluhrs, line 244: This should have worked. Unless DhsClientSim was reinstantiated for every step.

jluhrs (approved):
LGTM, but we should bump lucuma-core up in all instances, eventually.

rpiaggio:
> LGTM, but we should bump lucuma-core up in all instances, eventually.

Definitely agree.

modules/server_new/src/main/scala/observe/server/keywords/DhsClientHttp.scala
rpiaggio: It doesn't work when you restart the server.

PR #981, opened 2024-02-19
Single image deploy

rpiaggio:
This PR implements docker deployment via a single image for all environments and site, rather than having one image for each server.

The upside is that we can now deploy to testing or production the exact same image that we test in staging.

The downside is that now local configuration must be copied manually to each server.

Furthermore, this seems somewhat necessary to perform deployments via Dockerhub, where we seem to have a single artifact name available. We would have to discriminate server deploys via tags, which doesn't feel right.


jluhrs (approved):
LGTM

PR #1005, opened 2024-03-05
Scala 3.4 compliant syntax

rpiaggio:


modules/model/shared/src/main/scala/observe/common/FixedLengthBuffer.scala
```diff
@@ -44,13 +44,13 @@ object FixedLengthBuffer {
 
   def apply[A](maxLength: Int, initial: A*): Option[FixedLengthBuffer[A]] =
     (maxLength > 0 && maxLength >= initial.length)
-      .option(FixedLengthBufferImpl[A](maxLength, Chain(initial: _*)))
+      .option(FixedLengthBufferImpl[A](maxLength, Chain(initial*)))
```
cquiroz, line 47: C-flashback

cquiroz (approved):
LGTM

PR #1006, opened 2024-03-05
Bump versions

rpiaggio:


modules/server_new/src/main/scala/observe/server/odb/OdbProxy.scala
```diff
@@ -448,10 +442,12 @@ object OdbProxy {
         .map(RecordedStepId(_))
 
     private def recordDataset(stepId: RecordedStepId, fileId: ImageFileId): F[DatasetId] =
-      RecordDatasetMutation[F]
-        .execute(stepId.value, normalizeFilename(fileId.value))
-        .map(_.recordDataset.dataset.id)
-
+      Sync[F]
+        .delay(Dataset.Filename.parse(fileId.value).get)
```
rpiaggio, line 446: Should we scrap `ImageFileId` (which is a `NewType[String]`) and migrate to `Dataset.Filename` everywhere?

jluhrs (approved):
LGTM

PR #1023, opened 2024-03-14
Ported to coulomb

cquiroz:
This ended up being a largish PR. In here we are removing `squants` and replacing it with `coulomb` which will make it easier to move code to `lucuma-core`

In general we are converting:
* `Length` to `Quantity[Double, Millimeter]`
* `Time` to `TimeSpan`
* `Angle` to either lucuma `Angle` or `Quantity[Double, ArcSecond]`

A few extra tests were added to check for possible regressions


modules/server_new/src/main/scala/observe/server/altair/AltairControllerEpics.scala
```diff
@@ -26,23 +32,18 @@ import observe.server.tcs.Gaos.ResumeCondition.OiOn
...
 import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoUnit
 import scala.annotation.unused
 
 object AltairControllerEpics {
-  final case class EpicsAltairConfig(
-    currentMatrixCoords:  (Length, Length),
-    preparedMatrixCoords: (Length, Length),
+  case class EpicsAltairConfig(
+    currentMatrixCoords:  (Quantity[Double, Millimeter], Quantity[Double, Millimeter]),
```
rpiaggio, line 45: `Length` seems much more legible than the `Quantity[...]`. What do you thinking of using an (opaque?) type alias?

cquiroz: Not sure, at any rate it would be `LengthInMm`

modules/server_new/src/main/scala/observe/server/altair/AltairControllerEpics.scala
```diff
@@ -524,24 +534,24 @@ object AltairControllerEpics {
...
       .delay(LocalDate.now)
       .flatMap(date =>
         (epicsTcs.aoStatistics.setTriggerTimeInterval(0.0) *>
-          epicsTcs.aoStatistics.setInterval(expTime.toSeconds) *>
+          epicsTcs.aoStatistics.setInterval(expTime.toSeconds.toDouble) *>
           epicsTcs.aoStatistics.setSamples(1) *>
           epicsTcs.aoStatistics.setFileName(
             "aostats" + date.format(DateTimeFormatter.ofPattern("yyyyMMdd"))
-          )).whenA(expTime > 5.seconds && cfg =!= AltairOff)
+          )).whenA(TimeSpan.fromSeconds(5).exists(expTime > _) && cfg =!= AltairOff)
```
rpiaggio, line 545:
```suggestion
          )).whenA(expTime.toSeconds > 5) && cfg =!= AltairOff)
```

rpiaggio:
> Angle to either lucuma Angle or Quantity[Double, ArcSecond]

Maybe we can consider migrating to `opaque type Angle = Quantity[Long, MicroArcSecond]`?

modules/model/shared/src/main/scala/observe/model/package.scala
```diff
@@ -28,24 +25,12 @@ val CalibrationQueueName: String = "Calibration Queue"
...
 //      case Instrument.F2    => true
     case Instrument.GmosSouth => true
     case Instrument.GmosNorth => true
     case _                    => false
-//      case Instrument.Nifs  => true
-//      case Instrument.Niri  => true
-//      case Instrument.Gnirs => true
-//      case Instrument.Gsaoi => false
-//      case Instrument.Gpi   => true
-//      case Instrument.Ghost => false
```
jluhrs, line 48: I left those there, to restore them later as we implement support for more instruments.

cquiroz:
> > Angle to either lucuma Angle or Quantity[Double, ArcSecond]
> 
> Maybe we can consider migrating to `opaque type Angle = Quantity[Long, MicroArcSecond]`?

We already have an `Angle` type on `lucuma-core`

rpiaggio:
> We already have an `Angle` type on `lucuma-core`

Yes, that's the one I was suggesting migrating to an opaque type over a quantity, instead of it being a wrapper of a `Long` (microarcsec).

modules/server_new/src/main/scala/observe/server/tcs/TcsKeywordReader.scala
```diff
@@ -383,49 +387,60 @@ object TcsKeywordsReaderEpics extends TcsKeywordDefaults {
...
-      )
-    }
+    // private def normalizeSignedAngle(v: Angle): Angle = {
+    //   val r = v.value % 360.0
+    //   Degrees(
+    //     if (r < -180.0) r + 360.0
+    //     else if (r >= 180.0) r - 360.0
+    //     else r
+    //   )
+    // }
```
rpiaggio, line 422: Commented code

modules/server_new/src/main/scala/observe/server/tcs/TcsNorth.scala
```diff
@@ -214,8 +216,8 @@ object TcsNorth {
     val offset =
       StepConfig.science.andThen(StepConfig.Science.offset).getOption(stepConfig).map { o =>
         InstrumentOffset(
-          OffsetP(o.p.toAngle.toDoubleDegrees.degrees),
-          OffsetQ(o.q.toAngle.toDoubleDegrees.degrees)
+          OffsetP(Angle.signedDecimalArcseconds.get(o.p.toAngle).toDouble.withUnit[ArcSecond]),
```
rpiaggio, line 219: Maybe provide an `Angle` constructor for `Offset(P|Q)`?

BTW, could we use the `Offset.P` and `Offset.Q` types from `lucuma-core`?

modules/server_new/src/main/scala/observe/server/tcs/TcsNorthControllerEpicsAo.scala
```diff
@@ -243,13 +243,13 @@ object TcsNorthControllerEpicsAo {
       guiderCurrent:   GuiderConfig,
       tcsGuideDemand:  TelescopeGuideConfig,
       guiderDemand:    GuiderConfig,
-      distanceSquared: Option[Area],
-      threshold:       Option[Length]
+      distanceSquared: Option[Quantity[Double, SquaredMillis]],
```
rpiaggio, line 246: Same suggestion for `Area` as for `Length`.

modules/server_new/src/main/scala/observe/server/tcs/TcsNorth.scala
cquiroz: I did it but there is an issue, The angles get involved in some calculations that need more precision than what Angle provides, breaking some of the unit test. Since these are values going to the hardware I think it should be accepted thus I reverted to `Quantity[Double, arcSecond]` and it got the unit tests working

jluhrs (approved):
LGTM

modules/server_new/src/main/scala/observe/server/tcs/package.scala
```diff
@@ -1,78 +1,100 @@
...
+val agTimeout: TimeSpan  = TimeSpan.unsafeFromDuration(90, ChronoUnit.SECONDS)
 
-    implicit class LengthOps(l: Length) {
-      def times(fps: FocalPlaneScale): Angle = fps * l
-      def *(fps:     FocalPlaneScale): Angle = times(fps)
-    }
+object FocalPlaneScale:
+  extension (l: Quantity[Double, Millimeter])
+    def times(fps: FocalPlaneScale): Quantity[Double, ArcSecond] = fps * l
+    def **(fps:    FocalPlaneScale): Quantity[Double, ArcSecond] = times(fps)
```
rpiaggio, line 49: I'd discourage using `**` as multiplication. It can be confusing since it's exponentiation in some languages (eg: Python, JS).

cquiroz: right, the issue is that Quantity has already a `*` operator and the compiler would be all confused. What about `:*`?

modules/server_new/src/main/scala/observe/server/altair/AltairControllerEpics.scala
cquiroz: @jluhrs What do you think about using aliases for these types? They kind of confuse me but could be an option

modules/server_new/src/main/scala/observe/server/tcs/TcsNorth.scala
rpiaggio: `Angle` has µarcsec precision.

cquiroz: Instrument offsets are very small and they get multiplied for a small factor in process, In the case of an error I had 0.8 mas which was truncated to 0 making it not very good for mathematical operations

modules/server_new/src/main/scala/observe/server/altair/AltairControllerEpics.scala
rpiaggio: > Not sure, at any rate it would be `LengthInMm`

It would hide the internal representation and could have methods to extract in/construct from specific units (`toMillimeters`, `fromMillimeters`, etc.). Similar to how `Angle` behaves.

rpiaggio (approved):
LGTM

PR #1032, opened 2024-03-18
Support accepting guide configuration from an external source

cquiroz:
In this PR we remove a lot of classes that are now on `lucuma-core`. We also enable again sending guide configuration from an external entity, in this case navigate


jluhrs (approved):
LGTM

PR #1036, opened 2024-03-19
Fixed some bugs related to atoms

jluhrs:
Two bugs fixed:

- Record atom mutation was called always with GmosNorth instead of the actual instrument.
- A Step execution could be stopped after the exposure is completed, but before the last events are sent, which makes the step uncompleted and forces it to be run again.


modules/server_new/src/main/scala/observe/server/SeqTranslate.scala
```diff
@@ -276,7 +278,7 @@ object SeqTranslate {
        b.map(x =>
          SequenceGen(
            sequence,
-           Instrument.GmosNorth,
+           insSpec.instrument,
```
jluhrs, line 281: What was I thinking ??

rpiaggio (approved):
LGTM

PR #1046, opened 2024-03-21
Send step abort event.

jluhrs:
Two changes:
 - A Step Abort event is now sent to the ODB whenever the execution is stopped because of a fault.
 - Allow to record a new Step even when the current one does not complete normally.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/odb/IdTrackerOps.scala
+++ b/modules/server_new/src/main/scala/observe/server/odb/IdTrackerOps.scala
@@ -68,12 +68,7 @@ trait IdTrackerOps[F[_]: MonadThrow](idTracker: Ref[F, ObsRecordedIds]):
         .at(obsId)
         .some
         .andThen(RecordedVisit.step)
-        .modify:
-          case Some(staleStepId) if stepId.isDefined =>
-            throw ObserveFailure.Unexpected:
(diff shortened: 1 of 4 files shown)


modules/server_new/src/main/scala/observe/server/odb/IdTrackerOps.scala
```diff
@@ -68,12 +68,7 @@ trait IdTrackerOps[F[_]: MonadThrow](idTracker: Ref[F, ObsRecordedIds]):
         .at(obsId)
         .some
         .andThen(RecordedVisit.step)
-        .modify:
-          case Some(staleStepId) if stepId.isDefined =>
-            throw ObserveFailure.Unexpected:
-              s"Attempted to set current stepId for [$obsId] when it was already set. " +
-                s"Existing value [$staleStepId], new attempted value [${stepId.get}]"
-          case _                                     => stepId.map(RecordedStep(_))
+        .replace(stepId.map(RecordedStep(_)))
```
rpiaggio, line 71: Should we keep the safeguard and unset the current Step.Id on failure?

jluhrs: I do unset the current Step.Id on failure, so yes, it should work even with the safeguard in place.

PR #1059, opened 2024-03-29
Remove legacy event layer

rpiaggio:
In seqexec, the server sent `ObserveEvent`s to the clients. Now in observe, those were funneled through a new `ClientEvent` layer.

This PR simplifies the logic by removing the `ObserveEvent` layer completely, having `ObserveEngine` produce `ClientEvent`s directly.

There's also new logic (via `TargetedClientEvent`) to indicate whether an event should be broadcast to all clients or to a single one.


modules/web/client/src/main/scala/observe/ui/components/services/ServerEventHandler.scala
```diff
@@ -43,6 +43,8 @@ trait ServerEventHandler:
   )(using Logger[IO]): IO[Unit] =
     val asyncRootModel = rootModelData.async
     event match
+      case ClientEvent.BaDum                                                     =>
```
cquiroz, line 46: `BaDum`?

cquiroz (approved):
Looks great
This is leftover of somee  of the restrictions present on the old seqexec where  we had to use legacy java code that couldn't  be used on the frontend

This is not needed anymore in gpp

PR #1063, opened 2024-04-01
Fixed encoding of GMOS grating order

jluhrs:
Observe was writing the order name (Zero, One), while the instrument expects the number as string ("0", "1")


toddburnside (approved):
LGTM

PR #1067, opened 2024-04-03
Correctly update steps when loading new atom.

jluhrs:


modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -1860,7 +1860,7 @@ object ObserveEngine {
                   .atSequence[F](obsId)
                   .modify { seq =>
                     val sg: SequenceData[F] = seq.focus(_.seqGen.nextAtom).replace(atm)
-                    seq
+                    sg
```
jluhrs, line 1863: And this was the whole problem. `sg` has the new `Atom`, `seq` was the old one.

rpiaggio (approved):
LGTM

PR #1072, opened 2024-04-04
Keep UI updated when step/atom is completed

rpiaggio:
In this PR:
- When a step completes, visits are refreshed (only from the last loaded visit), to show the newly completed step.
- When an atom completes, it is removed from the `possibleFuture`.

`ServerEventHandler`  was reworked to receive modifier callbacks instead of `View`s, since the `View`s it could receive were stale. In other words, it cannot read values, only modify them.

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/arb/ArbClientEvent.scala
+++ b/modules/model/shared/src/main/scala/observe/model/arb/ArbClientEvent.scala
@@ -4,7 +4,9 @@
 package observe.model.arb
 
 import lucuma.core.enums.Instrument
+import lucuma.core.enums.SequenceType
 import lucuma.core.model.Observation
+import lucuma.core.model.sequence.Atom
 import lucuma.core.model.sequence.Step
 import lucuma.core.util.arb.ArbEnumerated.given
 import lucuma.core.util.arb.ArbGid.given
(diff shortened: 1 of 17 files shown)


modules/model/shared/src/main/scala/observe/model/arb/ArbClientEvent.scala
```diff
@@ -98,32 +100,55 @@ trait ArbClientEvent:
...
   given Cogen[ClientEvent] =
     Cogen[Either[
       Unit,
       Either[
         ClientEvent.InitialEvent,
-        Either[ClientEvent.ObserveState, Either[
-          ClientEvent.SingleActionEvent,
-          Either[ClientEvent.ChecksOverrideEvent, ClientEvent.ProgressEvent]
-        ]]
+        Either[
```
cquiroz, line 129: nice type

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -169,7 +175,18 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
         )
-      .render: (props, resize, cols, _, _, _, table) =>
+      .useContext(ODBQueryApi.ctx)
+      .useEffectWithDepsBy(
+        (props, _, _, _, _, _, _, _) => // If the list of current steps changes, reload last visit
+          props.currentAtomPendingSteps.map(_.id)
+      ): (_, _, _, _, _, _, _, odbQueryApi) =>
+        // TODO Maybe this should be done by ObservationSyncer. For that, we need to know there when a step
+        // has completed. Maybe we can add an ODB event in the future.
+        // ALSO TODO Put some state somewhere to indicate that the visits are reloading, new rows should
```
cquiroz, line 185: Good idea

cquiroz (approved):
LGTM

PR #1079, opened 2024-04-05
Filter observations by site in the UI

rpiaggio:


modules/model/shared/src/main/scala/observe/model/package.scala
```diff
@@ -38,6 +40,21 @@ extension (i: Instrument)
 //      case Instrument.Gpi   => true
 //      case Instrument.Ghost => false
 
+  def site: Option[Site] = i match
```
cquiroz, line 43: Does iit need to be an option? instruments always have a site?

rpiaggio: The `Instrument` enum has things like `Michelle`, `Visitor`, `Phoenix`, etc.

cquiroz: @swalker2m Why do we have those ancient instruments, are we planning to port the old ocs data?

cquiroz (approved):
LGTM

PR #1082, opened 2024-04-09
Auto-scroll to current step in sequence table

hugo-vrijswijk:


modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -128,6 +129,12 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
     extends SequenceTableDefs[D]:
   private type Props = SequenceTable[S, D]
 
+  private def scrollIfNeeded(element: dom.Element) = Callback {
```
cquiroz, line 132: Do we have this function in explore too? maybe eventually move it to lucuma-ui

hugo-vrijswijk: Yeah we have it in one place

cquiroz (approved):
LGTM

rpiaggio:
I think we may need to involve the virtualizer in this, like we do in the configuration table in explore. The step may not be rendered if it's beyond the current rows shown by the virtualizer.

Also, it may be nice to scroll to the next unexecuted step even if it's not selected.

PR #1083, opened 2024-04-09
Add logout button

hugo-vrijswijk:


modules/web/client/src/main/scala/observe/ui/Icons.scala
```diff
@@ -93,6 +93,10 @@ object Icons:
   @JSImport("@fortawesome/pro-solid-svg-icons", "faGears")
   private val faGears: FAIcon = js.native
 
+  @js.native
+  @JSImport("@fortawesome/pro-solid-svg-icons", "faRightFromBracket")
```
cquiroz, line 97: Is this the same as in explore?

hugo-vrijswijk: Yes

PR #1084, opened 2024-04-10
Add button to expand/collapse all visits

hugo-vrijswijk:


modules/web/client/src/main/scala/observe/ui/Icons.scala
```diff
@@ -156,11 +164,14 @@ object Icons:
...
     faFileCross,
     faGears,
+    faMinus,
     faMoon,
     faPause,
     faPlay,
+    faPlus,
     faRectangleList,
     faReply,
+    faSignOutAlt,
```
hugo-vrijswijk, line 174: I forgot to add this before, but the icon was still working. So I'm not sure what this list actually does

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -269,15 +273,31 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
-          rowMod = row => computeRowMods(row.original),
-          headerCellMod = computeHeaderCellMods,
-          cellMod = computeCellMods
+        React.Fragment(
+          // Render the expand all button in the ObsHeader
+          Option(dom.document.getElementById("sequence-table-expand-all")).map {
+            val (icon, label) =
+              if table.getIsAllRowsExpanded() then (Icons.Minus, "Collapse all visits")
+              else (Icons.Plus, "Expand all steps")
+            ReactPortal(
```
hugo-vrijswijk, line 282: This feels a bit hacky. The "portal" renders this component (the button) in a different node (from the getElementById), which is actually in the ObsHeader.scala. But I couldn't think of a clean way to get the button to render there (since you need the `table` object)

modules/web/client/src/main/scala/observe/ui/Icons.scala
cquiroz: Strange that should be needed to actually load the esm module to get the icon

modules/web/client/src/main/scala/observe/ui/components/sequence/ObsHeader.scala
```diff
@@ -42,6 +42,7 @@ object ObsHeader:
           s"${props.observation.title} [${props.observation.obsId}]"
         ),
         <.div(ObserveStyles.ObsSummaryDetails)(
+          <.span(^.id := "sequence-table-expand-all"),
```
cquiroz, line 45: What is this for?

cquiroz (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/components/sequence/ObsHeader.scala
hugo-vrijswijk: This is the dom element the react portal attaches to from the SequenceTable

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -269,15 +273,31 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
-          estimateSize = _ => 25.toPx,
-          overscan = 8,
-          containerRef = resize.ref,
-          tableMod = TagMod(tableStyle),
-          rowMod = row => computeRowMods(row.original),
-          headerCellMod = computeHeaderCellMods,
-          cellMod = computeCellMods
+        React.Fragment(
+          // Render the expand all button in the ObsHeader
+          Option(dom.document.getElementById("sequence-table-expand-all")).map {
```
rpiaggio, line 278: Can we use a `ref` for the DOM Node? We would have to define it in a common ancestor.

PR #1100, opened 2024-04-25
Backend interface for requesting next atom.

jluhrs:
This PR only implements the entry point for requesting more acquisition atoms. It does not include the actual code to load new atoms.
It also defines a new event that will eventually inform the client that the current atom run to completion. The event itself is not yet emitted in this version.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
+++ b/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
@@ -20,6 +20,7 @@ import lucuma.core.enums.Breakpoint
 import lucuma.core.enums.CloudExtinction
 import lucuma.core.enums.ImageQuality
 import lucuma.core.enums.Instrument
+import lucuma.core.enums.SequenceType
 import lucuma.core.enums.Site
 import lucuma.core.enums.SkyBackground
 import lucuma.core.enums.WaterVapor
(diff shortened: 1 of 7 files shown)


modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -91,6 +92,15 @@ trait ObserveEngine[F[_]] {
     runOverride: RunOverride
   ): F[Unit]
 
+  def nextAtom(
```
rpiaggio, line 95: May I suggest adding a verb since this seems to be an action method? For example, `loadNextAtom`

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
```diff
@@ -69,6 +69,12 @@ class ObserveCommandRoutes[F[_]: Async: Compression](
       ssoClient.require(req): user =>
         oe.requestCancelPause(obsId, obs, user) *> NoContent()
 
+    case req @ POST -> Root / ObsIdVar(obsId) / ClientIDVar(clientId) / "nextAtom" /
```
rpiaggio, line 72: Will this load and start the nextAtom?

rpiaggio (approved):
LGTM, thank you. With this I can start on the UI.

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
jluhrs: Yes, that's the idea.

PR #1106, opened 2024-04-29
Fix duplicate expand buttons and extract observation execution component

hugo-vrijswijk:
Home.scala was the common ancestor between the React Portal source and target, so this also extracts a separate component that has a `useState` because I didn't want to have the `useState`in the Home.scala.

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
@@ -57,6 +57,7 @@ sealed trait SequenceTable[S, D <: DynamicConfig](
   def requests: ObservationRequests
   def isPreview: Boolean
   def flipBreakpoint: (Observation.Id, Step.Id, Breakpoint) => Callback
+  def setExpandedButton: VdomNode => Callback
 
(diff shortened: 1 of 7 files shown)


modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -284,21 +284,18 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
                 case _                                    =>
                   TagMod.empty
 
+        val (icon, label) =
+          if table.getIsAllRowsExpanded() then (Icons.Minus, "Collapse all visits")
+          else (Icons.Plus, "Expand all steps")
```
rpiaggio, line 289: `Expand all visits`?

rpiaggio (approved):
LGTM, thank you!

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
hugo-vrijswijk: 👍

PR #1115, opened 2024-05-03
Only expand/collapse visits in sequence table and add separate buttons

hugo-vrijswijk:
afbeelding


modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -284,16 +287,51 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
                 case _                                    =>
                   TagMod.empty
 
-        val (icon, label) =
-          if table.getIsAllRowsExpanded() then (Icons.Minus, "Collapse all visits")
-          else (Icons.Plus, "Expand all visits")
+        def modExpanded(
+          f: (
+            Row[Expandable[HeaderOrRow[SequenceIndexedRow[D]]]]
```
rpiaggio, line 292: There's a type alias defined `SequenceTableRowType = Expandable[HeaderOrRow[SequenceIndexedRow[D]]]` which can be used as a shorthand.

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -284,16 +287,51 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
+          ) => Boolean
+        ): Callback = Callback.suspend:
+          table.setExpanded:
```
rpiaggio, line 295: There's a `table.modExpanded` that could be used and have the row.id check here, instead of repeating it in both calls below.

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -284,16 +287,51 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
+            Expanded.Rows(
+              table
+                .getExpandedRowModel()
+                .rows
+                .map: row =>
+                  RowId(row.id) -> f(Row(row))
+                .toMap
+            )
+
+        val allVisitsAreExpanded = table
```
rpiaggio, line 305: Similarly, repetition could be avoided by extracting a method that checks if all visits have expanded == (true | false).

rpiaggio (approved):
LGTM, just a couple of suggestions to avoid code duplication.

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
hugo-vrijswijk: Thanks, typing out that long type made me need a vacation!

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
hugo-vrijswijk: I wasn't sure if we could use it, since `Expanded.AllRows` doesn't have any info about current rows. But it does work (at least on my machine).

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
hugo-vrijswijk: I've refactored it a little bit to reduce code reuse

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -227,7 +227,10 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
                 sequence.indexWhere(row => CurrentHeadersRowIds.contains_(getRowId(row).value)) - 1,
                 ScrollOptions
               )
-      .render: (props, resize, cols, _, _, _, table, _, virtualizerRef) =>
+      .useMemoBy((_, _, _, visits, _, _, _, _, _) => visits)((_, _, _, _, _, _, _, _, _) =>
```
rpiaggio, line 230: Let's use a regular `val` for the `id` set.

By using `useMemo` we are comparing the whole list of `Visit`s on each rerender, which is probably more expensive than just building the `id` set.

Other than this, still LGTM.

hugo-vrijswijk: Done, I'll merge this now

PR #1120, opened 2024-05-07
Avoid progress bar from going backwards

rpiaggio:
Due to the smoothing mechanism, the observation progress bar sometimes went a little beyond the actual value, which caused the bar to recede when the actual progress value was received from the server.

This PR avoids that by memoizing the last value and avoiding a new value smaller than the previous one.

The progress bar is still a bit jumpy, so the smoothing mechanism needs some more tweaking. That will be addressed in a future PR.


cquiroz (approved):
LGTM

PR #1122, opened 2024-05-07
Upgrade libs with the now required explicit circe codecs

rpiaggio:
The new version of circe requires explicit codecs, whereas previous versions would derive codecs for all referenced classes.

This means that we now have to save the instrument in `ObserveStep` in order to be able to decode the correct `DynamicConfig` (I have no idea what circe what doing with this before).

I created a wrapper `InstrumentDynamicConfig` solely for the purpose of transporting the generic `DynamicConfig`.

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/InstrumentDynamicConfig.scala
+++ b/modules/model/shared/src/main/scala/observe/model/InstrumentDynamicConfig.scala
@@ -1,0 +1,70 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package observe.model
+
+import cats.Eq
+import cats.syntax.either.*
+import io.circe.Decoder
+import io.circe.DecodingFailure
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.syntax.*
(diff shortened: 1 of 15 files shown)


modules/model/shared/src/main/scala/observe/model/InstrumentDynamicConfig.scala
```diff
@@ -0,0 +1,70 @@
...
+  def config: DynamicConfig
+
+  case GmosNorth(config: DynamicConfig.GmosNorth)
+      extends InstrumentDynamicConfig(Instrument.GmosNorth)
+
+  case GmosSouth(config: DynamicConfig.GmosSouth)
+      extends InstrumentDynamicConfig(Instrument.GmosSouth)
+
+object InstrumentDynamicConfig:
+  // object GmosNorth:
```
cquiroz, line 31: Commented code

modules/model/shared/src/main/scala/observe/model/ObserveStep.scala
```diff
@@ -27,10 +23,13 @@ import monocle.syntax.all.*
 import observe.model.dhs.*
 import observe.model.enums.*
 import observe.model.enums.PendingObserveCmd
+// import observe.model.codecs.given
```
cquiroz, line 26: Commented code

cquiroz (approved):
LGTM

PR #1123, opened 2024-05-08
Atom start/end invocations in ODBProxy

rpiaggio:
The actual record of the `StartAtom` event is already being executed as part of the `atomStart` method.

Invocation to `atomEnd` from the engine is pending.


cquiroz (approved):
LGTM

PR #1125, opened 2024-05-09
ApiClient in state instead of just a val

rpiaggio:
Having the `ApiClient` in state helps us force a rerender whenever the user token changes. This is necessary, since otherwise the rendered `onClick`s went stale when the token changed, causing an error when invoking the server.


cquiroz (approved):
LGTM

PR #1126, opened 2024-05-09
Refresh sequence after each completed step

rpiaggio:
This should happen after each atom, but that requires rearranging initialization. It will be addressed in a future PR.

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/services/ODBQueryApiImpl.scala
+++ b/modules/web/client/src/main/scala/observe/ui/services/ODBQueryApiImpl.scala
@@ -8,6 +8,7 @@ import cats.syntax.all.*
 import clue.*
 import crystal.ViewF
 import lucuma.schemas.ObservationDB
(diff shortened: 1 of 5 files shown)


modules/web/client/src/main/scala/observe/ui/services/ODBQueryApiImpl.scala
```diff
@@ -28,3 +29,24 @@ case class ODBQueryApiImpl(nighttimeObservation: ViewF[IO, Option[LoadedObservat
         .attempt
         .flatMap: visits =>
           loadedObs.mod(_.withVisits(visits))
+
+  override def refreshNighttimeSequence: IO[Unit] =
+    nighttimeObservation.toOptionView.fold(
+      Logger[IO].error("refreshNighttimeSequence with undefined loaded observation")
+    ): loadedObs =>
+      IO.println("hello") >>
```
cquiroz, line 37: hello

cquiroz (approved):
LGTM

PR #1127, opened 2024-05-09
Incremental visits query

rpiaggio:
When requerying visits for new entries, only request from the last visit already loaded (included, there can be new recorded atoms/steps in it), instead of reloading the whole list.


cquiroz (approved):
LGTM

PR #1128, opened 2024-05-10
Incremental visits query actually working

rpiaggio:
There was a bug in the previous implementation, where existing visits were deleted.


cquiroz (approved):
LGTM

PR #1129, opened 2024-05-10
Auto scroll past the visits on sequence load, using virtualizer

rpiaggio:
By using the virtualizer, we ensure the scroll works even when there are lots of visits and the current sequence wasn't rendered initially.


cquiroz (approved):
LGTM

PR #1137, opened 2024-05-19
Move execution state to table meta

rpiaggio:
This is now possible with the new react-table facade.

It prevents table components from remounting on each execution status update.

There are other migrations to the new table facade too.

Note: The progress bar lost its "smoothness". Its current code compensates for the the remounts, and the logic has to be simplified now. Will be addressed in a future PR.


cquiroz (approved):
LGTM
Will you apply these changes to explore too?

PR #1161, opened 2024-05-30
Bump versions

rpiaggio:
The observation title is now a `NonEmptyString`.

I propagated this change to `SequenceMetadata`. I did not propagate it to `obsId` in `ObsKeywordsReader`.


modules/model/shared/src/main/scala/observe/model/arb/ObserveModelArbitraries.scala
```diff
@@ -92,7 +94,7 @@ trait ObserveModelArbitraries {
     for {
       i <- arbitrary[Instrument]
       o <- arbitrary[Option[Observer]]
-      n <- Gen.alphaStr
+      n <- Gen.alphaStr.suchThat(_.nonEmpty).map(NonEmptyString.unsafeFrom)
```
cquiroz, line 97: Good one

cquiroz (approved):
LGTM

PR #1175, opened 2024-06-05
Edit datasets QA

rpiaggio:
https://github.com/gemini-hlsw/observe/assets/1895643/342db951-1881-43e4-b5be-e3d3be1e90ab

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
@@ -133,15 +133,16 @@ object MainApp extends ServerEventHandler:
   private def enforceStaffRole(ssoClient: SSOClient[IO]): IO[Option[UserVault]] =
     ssoClient.whoami
       .flatMap: userVault =>
-        if LinkingInfo.developmentMode then userVault.pure[IO]
-        else
-          userVault.map(_.user) match
-            case Some(StandardUser(_, role, other, _)) =>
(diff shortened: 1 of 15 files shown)


cquiroz:
beautiful

modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
```diff
@@ -133,15 +133,16 @@ object MainApp extends ServerEventHandler:
   private def enforceStaffRole(ssoClient: SSOClient[IO]): IO[Option[UserVault]] =
     ssoClient.whoami
       .flatMap: userVault =>
-        if LinkingInfo.developmentMode then userVault.pure[IO]
```
cquiroz, line 136: No need forr a dev mode?

cquiroz (approved):
LGTM looks great

modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
rpiaggio: WDYM?

rpiaggio: The logic has changed. Observe still lets you log in while in development with non-staff users. But if the user has the STAFF role, it will still attempt to switch to it. This is because the ODB checks the user is logged in with the STAFF role to allow setting the QA state of a dataset.

PR #1180, opened 2024-06-05
Autoscroll on load and notable status changes

rpiaggio:
Autoscroll sequence on load to:
1. Alert row, orElse
2. Running row, orElse
3. Next row to run

Autoscroll sequence on sequence start/request more rows/stop to:
1. Alert row, orElse
2. Running row (which is still set on failed/aborted).

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
@@ -9,6 +9,7 @@ import crystal.react.hooks.*
 import crystal.react.syntax.effect.*
 import eu.timepit.refined.types.numeric.NonNegInt
 import japgolly.scalajs.react.*
+import japgolly.scalajs.react.hooks.Hooks.UseRef
 import japgolly.scalajs.react.vdom.html_<^.*
 import lucuma.core.enums.Breakpoint
 import lucuma.core.enums.DatasetQaState
@@ -112,6 +113,12 @@ sealed trait SequenceTable[S, D <: DynamicConfig](
   // Alert position is right after currently executing atom.
   protected[sequence] lazy val alertPosition: NonNegInt =
     NonNegInt.unsafeFrom(currentAtomPendingSteps.length)
+
+  protected[sequence] lazy val runningStepId: Option[Step.Id] = executionState.runningStepId
+
+  protected[sequence] lazy val nextStepId: Option[Step.Id] =
+    currentAtomPendingSteps.headOption.map(_.id)
+
 case class GmosNorthSequenceTable(
   clientMode:           ClientMode,
   obsId:                Observation.Id,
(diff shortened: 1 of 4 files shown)


modules/web/client/src/main/webapp/styles/prime.scss
```diff
@@ -30,10 +30,7 @@
 
 body,
 body .p-component {
-  margin: 0;
```
rpiaggio, line 33: Unrelated. This was unnecessary and was interfering with the new primereact styles.

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -279,14 +299,30 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
+        val autoScrollCandidates: List[String] =
+          AlertRowId.toString +:
+            (props.runningStepId ++ props.nextStepId).map(_.toString).toList
+
+        // If sequence is not running, auto select next step.
+        Callback.when(props.runningStepId.isEmpty)(
+          props.nextStepId.map(props.setSelectedStepId).orEmpty
+        ) >>
+          scrollToRowId(virtualizerRef, table)(autoScrollCandidates)
+            .delayMs(1) // https://github.com/TanStack/virtual/issues/615
```
rpiaggio, line 312: `scrollToRowId` doesn't work `onMount` in the latest virtualizer version, unless `delayMs` is used.

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -164,8 +171,21 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
+  )(rowIdCandidates: List[String]): Callback =
+    virtualizerRef.get.flatMap: refOpt =>
+      Callback( // Auto scroll to running step or next step.
+        refOpt.map:
+          _.scrollToIndex(
+            table
+              .getRowModel()
+              .rows
+              .indexWhere(row => rowIdCandidates.contains_(row.id.value)) - 1,
+            ScrollOptions
```
cquiroz, line 186: What ScrollOptions do we use?

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
cquiroz: Interesting, even 1ms works?

cquiroz (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
rpiaggio:
```scala
  private val ScrollOptions =
    rawVirtual.mod
      .ScrollToOptions()
      .setBehavior(rawVirtual.mod.ScrollBehavior.smooth)
      .setAlign(rawVirtual.mod.ScrollAlignment.start)
```

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
rpiaggio: The goal is to just delay it to the next tick in the event loop. In JS it would even work with 0, but `delayMs` has an optimization where if it receives 0 it just skips the `setTimeout` and executes right away.

PR #1182, opened 2024-06-07
Show dataset links (if valid)

rpiaggio:


modules/web/client/src/main/scala/observe/ui/model/AppContext.scala
```diff
@@ -37,6 +39,8 @@ case class AppContext[F[_]](
   val logger:    Logger[F],
   val odbClient: WebSocketJSClient[F, ObservationDB]
 ):
+  given Client[F] = httpClient
```
cquiroz, line 42: Why not pass this as `using‘?

cquiroz (approved):
LGTM

modules/web/client/src/main/scala/observe/ui/model/AppContext.scala
rpiaggio: Hmmm... The site where this is constructed has multiple clients in scope. Maybe this shouldn't be implicit. In which case, we have to modify all the way down to lucuma-ui. I'll address this in a future PR.

PR #1275, opened 2024-07-31
Show program reference in observation table

rpiaggio:


modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
```diff
@@ -38,7 +39,8 @@ case class ObsSummary(
   attachmentIds:      SortedSet[ObsAttachment.Id],
   observingMode:      Option[ObservingMode],
   visualizationTime:  Option[Instant],
-  posAngleConstraint: PosAngleConstraint
+  posAngleConstraint: PosAngleConstraint,
+  programReference:   Option[ProgramReference]
```
cquiroz, line 43: We are showing the program ref but if not present the proposal ref. Should we do the same here?

rpiaggio: I considered that, but are we actually going to get `READY` observations in observe without a program ref?

cquiroz: Can PIs change the obs status?

cquiroz (approved):
LGTM

PR #1276, opened 2024-08-01
Show observation reference in table instead of program reference + observation id

rpiaggio:
image

### DIFF
--- a/modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
+++ b/modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
@@ -16,8 +16,8 @@ import lucuma.core.enums.Instrument
 import lucuma.core.model.ConstraintSet
 import lucuma.core.model.ObsAttachment
 import lucuma.core.model.Observation
+import lucuma.core.model.ObservationReference
 import lucuma.core.model.PosAngleConstraint
-import lucuma.core.model.ProgramReference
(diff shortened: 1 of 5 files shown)


modules/web/client-model/src/main/scala/observe/ui/model/ObsSummary.scala
```diff
@@ -82,8 +82,7 @@ object ObsSummary:
       observingMode      <- c.get[Option[ObservingMode]]("observingMode")
       visualizationTime  <- c.get[Option[Timestamp]]("visualizationTime")
       posAngleConstraint <- c.get[PosAngleConstraint]("posAngleConstraint")
-      programReference   <-
-        c.downField("program").downField("reference").get[Option[ProgramReference]]("label")
+      obsReference       <- c.downField("reference").get[Option[ObservationReference]]("label")
```
cquiroz, line 85: These may fail if `reference` is null

rpiaggio: Doesn't it return `None`? Anyway, this is defensive. Ready observations are expected to always have a `reference`.

rpiaggio: I can confirm that it returns `None` instead of failing.

cquiroz (approved):
LGTM

PR #1282, opened 2024-08-13
Several FITS keyword fixes

jluhrs:


modules/server_new/src/main/scala/observe/server/gws/package.scala
```diff
@@ -0,0 +1,22 @@
+package observe.server
+
+import coulomb.*
+import coulomb.define.*
+
+package object gws {
+
+  export coulomb.units.accepted.{Millibar, ctx_unit_Millibar}
+
+  final type Bar
```
cquiroz, line 10: We have some of these in core. I suggest we move these there

modules/server_new/src/main/scala/observe/server/odb/OdbProxy.scala
```diff
@@ -140,8 +143,15 @@ object OdbProxy {
...
       false.pure[F]
 
-    override def datasetStartExposure(obsId: Observation.Id, fileId: ImageFileId): F[Boolean] =
-      true.pure[F]
+    override def datasetStartExposure(
+      obsId:  Observation.Id,
+      fileId: ImageFileId
+    ): F[RecordDatasetMutation.Data.RecordDataset.Dataset] =
+      Sync[F]
+        .delay(scala.util.Random.between(1L, Long.MaxValue))
```
cquiroz, line 151: You can use
https://typelevel.org/cats-effect/docs/std/random

modules/server_new/src/main/scala/observe/server/gws/GwsKeywordReader.scala
```diff
@@ -0,0 +1,104 @@
...
+import observe.server.EpicsHealth
+import observe.server.keywords.*
+
+object DewPoint extends NewType[Temperature[Double, Celsius]]
+type DewPoint = DewPoint.Type
+
+trait GwsKeywordReader[F[_]] {
+  def health: F[EpicsHealth]
+
+  def temperature: F[Temperature[Double, Celsius]]
```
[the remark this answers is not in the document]
jluhrs: No, this is the ambient temperature. The next function is the one used to read the dew point.

modules/server_new/src/main/scala/observe/server/gws/package.scala
jluhrs: Where else would those be used?
I guess that some reporting tools will show weather conditions

modules/server_new/src/main/scala/observe/server/odb/OdbProxy.scala
jluhrs: I tried after reading this comment, but for this particular use the current solution is simpler.

cquiroz (approved):
LGTM

PR #1323, opened 2024-09-09
Updated scala and libraries.

jluhrs:


modules/engine/src/main/scala/observe/engine/Handle.scala
```diff
@@ -102,18 +102,18 @@ object Handle {
   }
 
   extension [F[_]: Functor, D, A](self: StateT[F, D, A])
-    def toHandle[V]: Handle[F, D, V, A] = Handle[F, D, V, A](self.map((_, None)))
+    def toHandleT[V]: Handle[F, D, V, A] = Handle[F, D, V, A](self.map((_, None)))
```
cquiroz, line 105: Why the rename?

cquiroz (approved):
LGTM

modules/server_new/src/clue/scala/observe/common/ObsQueriesGQL.scala
```diff
@@ -285,7 +285,6 @@ object ObsQueriesGQL:
     val document = """
       subscription {
         observationEdit(programId:"p-2") {
-          id
```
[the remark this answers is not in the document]
jluhrs: The issue here is that the filed is deprecated

modules/engine/src/main/scala/observe/engine/Handle.scala
jluhrs: The name collided in one use. It worked before because I used and alias in an import, but for some reason now the import triggers an "unused" warning.

PR #1392, opened 2024-10-28
Workaround for bug on odb call

cquiroz:


modules/server_new/src/clue/scala/observe/common/ObsQueriesGQL.scala
```diff
@@ -363,10 +363,7 @@ object ObsQueriesGQL:
               label
               observation {
                 label
-                program {
```
cquiroz, line 366: This call is throwing an assertion on the odb but I'm unable to reproduce it in a test

modules/server_new/src/main/scala/observe/server/keywords/StandardHeader.scala
```diff
@@ -73,7 +73,7 @@ class StandardHeader[F[_]: Sync: Logger](
       buildString(obsObject, KeywordName.OBJECT),
       buildString(obsReader.obsType, KeywordName.OBSTYPE),
       buildString(obsReader.obsClass, KeywordName.OBSCLASS),
-      buildString(dataset.map(_.observation.program.label.label).getOrElse("").pure[F],
+      buildString(dataset.map(_.observation.label.programReference.label).getOrElse("").pure[F],
```
rpiaggio, line 76: Does this return the same value? Maybe we can leave it like this.

rpiaggio (approved):
LGTM

modules/server_new/src/main/scala/observe/server/keywords/StandardHeader.scala
cquiroz: it should be the same

PR #1400, opened 2024-10-30
Migrate to latest ODB changes: workflows and offsets in every sequence step

rpiaggio:


modules/server_new/src/clue/scala/observe/common/ObsQueriesGQL.scala
```diff
@@ -17,9 +17,12 @@ object ObsQueriesGQL:
 
   @GraphQL
   trait ActiveObservationIdsQuery extends GraphQLOperation[ObservationDB]:
+    // TODO The ODB API doesn't provide a way to filter ready observations,
+    // so we filter by accepted proposals for now.
+    // Revise this when the API supports it OR we start getting obersvations from the scheduler.
```
cquiroz, line 22: Wouldnt the obs workflow have a state saying we are ready?

modules/server_new/src/main/scala/observe/server/tcs/TcsSouth.scala
```diff
@@ -223,48 +224,30 @@ object TcsSouth {
...
-    )
-    val gwod3  = none.flatMap(_ =>
-      StepConfig.science.andThen(StepConfig.Science.guiding).getOption(stepConfig)
-    )
-    val gwod4  = none.flatMap(_ =>
-      StepConfig.science.andThen(StepConfig.Science.guiding).getOption(stepConfig)
-    )
+    val guiding: StepGuideState = telescopeConfig.guiding
+
+    val gwp1   = none.map(_ => guiding)
```
cquiroz, line 236: Wouldn't the `map` never be called?

cquiroz (approved):
LGTM

modules/server_new/src/clue/scala/observe/common/ObsQueriesGQL.scala
rpiaggio: Yes, but:
1) There's no way to set an observation as `Ready` yet.
2) There's no way to filter `Ready` observations.

modules/server_new/src/main/scala/observe/server/tcs/TcsSouth.scala
rpiaggio: Exactly. I don't know why code was this way, I supposed it is a reminder for when we work on the other instruments. I tried to respect that.

cquiroz: @jluhrs why's this?

PR #1408, opened 2024-11-05
SC-3912 Produced percentile keywords from new constraint types.

jluhrs:
What used to be a simple read of a value now uses conversion tables and values read from TCS, making it more complicated.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
+++ b/modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
@@ -3,54 +3,153 @@
 
 package observe.server.keywords
 
-import lucuma.core.enums.{Site, SkyBackground, WaterVapor}
+import cats.data.NonEmptyList
+import cats.syntax.all.*
+import lucuma.core.enums.CloudExtinction
+import lucuma.core.enums.ImageQuality
+import lucuma.core.enums.Site
+import lucuma.core.enums.SkyBackground
+import lucuma.core.enums.WaterVapor
+import lucuma.core.math.Angle
+import lucuma.core.math.Wavelength
 import lucuma.core.model.ConstraintSet
 import observe.model.Conditions
 
+import scala.annotation.tailrec
+import scala.math.Ordering.Implicits.infixOrderingOps
+import scala.math.Pi
+import scala.math.abs
+import scala.math.pow
+import scala.math.sin
+
 object ConditionOps {
-  
+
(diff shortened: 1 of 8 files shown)


modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
```diff
@@ -3,54 +3,153 @@
...
-        case WaterVapor.VeryDry => 1.0
-        case WaterVapor.Dry => 1.6
-        case WaterVapor.Median => 3.0
-        case WaterVapor.Wet => 5.0
-      case Site.GS => wv match
-        case WaterVapor.VeryDry => 2.3
-        case WaterVapor.Dry => 4.3
-        case WaterVapor.Median => 7.6
-        case WaterVapor.Wet => 10.0
+      case Site.GN =>
```
cquiroz, line 29: Do you think we should put things like this on db? Such that we can change them without rebuilding?

modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
```diff
@@ -3,54 +3,153 @@
...
+          NonEmptyList.of(0.35, 0.50, 0.75, 1.25),
+          NonEmptyList.of(0.35, 0.50, 0.70, 1.15),
+          NonEmptyList.of(0.31, 0.37, 0.45, 0.75)
+        )
+        .map(_.zip(NonEmptyList.of(20, 70, 85, 100)))
+    )
+
+  extension (iq: ImageQuality) {
+    def toPercentile(wv: Wavelength, elevation: Angle): Int = {
+      // Pickering approximation (thanks Andy Stephens)
```
cquiroz, line 102: What is pickering?
It seems this should go to `lucuma-core`

cquiroz (approved):
LGTM

modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
jluhrs: Pickering is a name. Ref: https://en.wikipedia.org/wiki/Air_mass_(astronomy)#Interpolative_formulas

modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
jluhrs: I don't know how frequently those values change. It is a question fro Andy.

PR #1437, opened 2024-11-28
Notify user of errors in engine

rpiaggio:
Notification events from the engine now make it all the way to the clients, where they are displayed in a sticky toast.


cquiroz (approved):
LGTM

PR #1438, opened 2024-11-29
Make observe reload the current atom at the end of each step

cquiroz:
This is a proposal of  how to solve the fact that atoms can change. Currently  the observe engine loads an atom and creates a local representation, if the atom changes that local representation needs to be updated.

This is to some extent  similar to loading the next atom thus the code has some commonality with loading the next atom.

In this PR at the end of executing a step we do a reload without emitting events (unlike loading the next atom)

Below is a video where a step is marked as failed thus forcing the odb to produce a new step. The exposure times have been reduced to make the example go faster. Note thaat we start with 3 steps but  at the end we execute 4 as the  first failed.

https://github.com/user-attachments/assets/1d267036-31fe-464b-9bfc-4c6839f8901b

Please check it and verify if this approach seems correct. Further testing is needed

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
+++ b/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
@@ -54,6 +54,7 @@ import observe.model.enums.PendingObserveCmd.*
 import observe.model.enums.Resource
 import observe.model.enums.RunOverride
 import observe.model.events.*
+import observe.server.SequenceGen.AtomGen
 import observe.server.events.*
 import observe.server.odb.OdbProxy
 import org.typelevel.log4cats.Logger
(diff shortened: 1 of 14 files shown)


rpiaggio (approved):
LGTM, but I can't say I fully understand the innards of the engine.

modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -70,8 +71,6 @@ trait ObserveEngine[F[_]] {
 
   val systems: Systems[F]
 
-  def sync(seqId: Observation.Id): F[Unit]
```
jluhrs, line 73: Why remove `sync` ?

jluhrs (approved):
LGTM

PR #1461, opened 2024-12-06
Build for linux/amd64, useful for local deployments

cquiroz:


rpiaggio:
Do we want this on the deployments that go to testing and production or only for the ones on docker?

cquiroz:
do we have a scenario wher we want another platform?

rpiaggio:
> do we have a scenario wher we want another platform?

🤷  We'll find out eventually.

rpiaggio (approved):
LGTM

PR #1468, opened 2024-12-09
Remove old client

cquiroz:
We have the old seqexec client code cluttering the project making search more difficult

We are probably not going to reuse any piece of that so I'd just remove it


rpiaggio (approved):
LGTM

PR #1475, opened 2024-12-10
Use odb breakpoints when creating the engine steps

cquiroz:


jluhrs:
So, this is all only to remove the default value for the breakpoint, and force it to specify one every time?

cquiroz:
Yes, codewise is not a big change, but it would reflect the breakpoints as set by the odb

rpiaggio (approved):
LGTM

PR #1477, opened 2024-12-11
Don't load obs attachments

cquiroz:
I don't think we are using them


rpiaggio:
The inception document mentions that finder charts should be accessible from observe, so I added the attachment ids foreseeing that.

PR #1480, opened 2024-12-13
Remove unused prism

cquiroz:
This prism is unlawful and it is not in used. It was used on the seqexec to parse ocs string-based model


rpiaggio (approved):
LGTM

PR #1484, opened 2024-12-16
Use the site to generate filenames

cquiroz:
When simulating the file names are always set as Gemini South which is confusing when simulating at GN.

This PR uses the server site configuration to decide


rpiaggio (approved):
LGTM

PR #1485, opened 2024-12-16
Hide buttons to collapse/hide visits if there are no visits

cquiroz:


modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
```diff
@@ -434,20 +434,22 @@ private sealed trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq]
...
+                onClick = collapseVisits
+              ).mini.compact,
+              Button(
+                icon = Icons.Plus,
+                label = "Expand all visits",
+                disabled = allVisitsAreExpanded,
+                onClick = expandVisits
+              ).mini.compact
+            )
+          } else EmptyVdom,
```
rpiaggio, line 452:
```suggestion
            <.div(ObserveStyles.SequenceTableExpandButton)(
              Button(
                icon = Icons.Minus,
                label = "Collapse all visits",
                disabled = allVisitsAreCollapsed,
                onClick = collapseVisits
              ).mini.compact,
              Button(
                icon = Icons.Plus,
                label = "Expand all visits",
                disabled = allVisitsAreExpanded,
                onClick = expandVisits
              ).mini.compact
            ).when(visitIds.nonEmpty),
```

rpiaggio (approved):
LGTM, with a more idiomatic suggestion

modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTable.scala
cquiroz: I tried that but `when` doesn't work on this context

PR #1500, opened 2024-12-18
Show obsid when obs reference is none

rpiaggio:


modules/web/client/src/main/scala/observe/ui/components/queue/SessionQueue.scala
```diff
@@ -195,10 +196,10 @@ object SessionQueue:
     ),
     ColDef(
       ObsRefColumnId,
-      _.obsReference,
+      obs => obs.obsReference.toRight((obs.programId, obs.obsId)),
```
cquiroz, line 199: Right, so we use the reference if available and default to progid/obsid if not
A question is whether we should load obs without a reference, I guess it will be needed for engineering?

cquiroz:
I guess a related question is whether we are writing the reference to keywords.

cquiroz (approved):
LGTM

PR #1501, opened 2024-12-18
Use stepAbort for aborting a step

cquiroz:
After aborting we cannot resume sequences. This was traced to the odb proxy which forgets the visit when doing `obsAbort`. It is unclear whats the role of `obsAbort` so it is removed and we use `steAbort` instead.

I wish we could make a test case but I couldn't figure out how to do it


rpiaggio (approved):
LGTM

PR #1520, opened 2024-12-23
Bump libraries, scala 3.6 migration

rpiaggio:


modules/model/shared/src/main/scala/observe/model/package.scala
```diff
@@ -54,26 +51,13 @@ extension (i: Instrument)
     // None
     case _                     => none
 
-// Resources come before Instruments
-given Order[Resource | Instrument] = Order.from:
```
cquiroz, line 58: Is this because they are `Enumerated`?

cquiroz (approved):
LGTM

modules/model/shared/src/main/scala/observe/model/package.scala
rpiaggio: Yes, `Enumerated` provides an `Order` already.

PR #1525, opened 2024-12-24
Show current aborted step as idle

rpiaggio:
Aborted steps are already shown as such in the visit record.

The current row of an aborted step should be shown as idle, without further styling.

Before:
image

After:
image


cquiroz (approved):
LGTM

PR #1559, opened 2025-01-13
Fixed calculation of IQ percentile values.

jluhrs:


modules/server_new/src/test/scala/observe/server/keywords/ConditionOpsSuite.scala
```diff
@@ -0,0 +1,46 @@
...
+import lucuma.core.math.Angle
+import lucuma.core.math.Wavelength
+import munit.FunSuite
+
+import ConditionOps.*
+
+class ConditionOpsSuite extends FunSuite {
+
+  test("ConditionOps calculate image quality percentile from double value") {
+    assertEquals(
```
jluhrs, line 16: This was scalafmt doing.

modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
```diff
@@ -101,8 +101,8 @@ object ConditionOps {
     def toPercentile(wv: Wavelength, elevation: Angle): Int = {
       // Pickering approximation (thanks Andy Stephens)
       val airmass         = 1.0 / sin(
-        (90.0 - elevation.toDoubleDegrees + 244.0 / (165.0 + 47.0 * pow(
```
jluhrs, line 104: In my defense, the original formula has the expression `(90.0 - z)`. Then again, that `z` is not elevation, but zenith angle.

modules/server_new/src/main/scala/observe/server/keywords/ConditionOps.scala
```diff
@@ -101,8 +101,8 @@ object ConditionOps {
     def toPercentile(wv: Wavelength, elevation: Angle): Int = {
```
cquiroz, line 101: Should these go to lucuma-core?

cquiroz (approved):
LGTM

PR #1607, opened 2025-01-28
SC-4148 Reload atom every time the execution starts.

jluhrs:


modules/engine/src/main/scala/observe/engine/Engine.scala
```diff
@@ -51,11 +51,14 @@ class Engine[F[_]: MonadThrow: Logger, S, U] private (
...
-            Sequence.State.status.replace(SequenceState.Running.Init)(
+            Sequence.State.status.replace(
+              SequenceState.Running(false, false, waitingNextAtom = true, isStarting = true)
+            )(
               seq.rollback
             )
           ) *>
-            send(Event.executing(id))
+            send(modifyState(atomReload(this, id)))
+          // send(Event.executing(id))
```
cquiroz, line 61: Commented line

modules/engine/src/main/scala/observe/engine/Engine.scala
```diff
@@ -201,13 +213,15 @@ class Engine[F[_]: MonadThrow: Logger, S, U] private (
                 case Sequence.State.Final[F](_, _) => send(finished(id))
                 // Execution completed. Check breakpoint here
                 case _                             =>
-                  if (seq.getCurrentBreakpoint) {
+                  if (!isStarting && seq.getCurrentBreakpoint) {
                     switch(id)(SequenceState.Idle) *> send(breakpointReached(id))
                   } else
-                    switch(id)(SequenceState.Running(false, false, false)) *> send(executing(id))
+                    switch(id)(SequenceState.Running(false, false, false, false)) *> send(
```
cquiroz, line 219: Shouldn't we do an `atomReload` when `isStarting` is true?

cquiroz (commented):
Great idea to store `isStarting` no the internal state.
It isn't clear to me though where is the atom being reloaded?

modules/engine/src/main/scala/observe/engine/Engine.scala
jluhrs: No. `isStarting` is used to disable the breakpoint check in `atomReload`.

jluhrs:
How it it work: instead of scheduling the execution of the sequence, `start` schedules the call to `atomReload`, which in turn will schedule the execution of the sequence after the atom is reloaded.

cquiroz:
I see, thanks for the clarificiation

cquiroz (approved):
LGTM

PR #1616, opened 2025-02-05
Catch all query

cquiroz:


toddburnside (approved):
👍

modules/web/client/src/clue/scala/observe/queries/ObsQueriesGQL.scala
```diff
@@ -17,7 +17,7 @@ object ObsQueriesGQL {
     // Revise this when the API supports it OR we start getting obersvations from the scheduler.
     val document = s"""
       query {
-        observations(WHERE: {program: {proposalStatus: {EQ: ACCEPTED}}}) {
+        observations() {
```
swalker2m, line 20: Eventually something like:

```
WHERE: {
  program: {
    OR: [
      { proposalStatus: { EQ: ACCEPTED } },
      { type: { NEQ: SCIENCE } }
    ]
  }
}
```
perhaps.

toddburnside: Close. But I think we only want ENGINEERING and CALIBRATION programs, not LIBRARY, etc. I'll put in a PR.

PR #1617, opened 2025-02-05
Limit observations to ACCEPTED, or ENGINEERING or CALIBRATION programs

toddburnside:
A similar WHERE should be used in the query in `Observation.ts` in navigate-ui.

Note, there is also an `observationsByWorkflowState` which would allow filtering by observation workflow state in addition to the WHERE used here. But, it is only available for service users, which I think means it can't be used here?

UPDATE: Rob made `observationsByWorkflowState` available to staff users. @hugo-vrijswijk is creating a PR for navigate-ui using that query. The exact query should be discussed in that PR and I will duplicate it here.

Hugo's PR: https://github.com/gemini-hlsw/navigate-ui/pull/403

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/extensions.scala
+++ b/modules/model/shared/src/main/scala/observe/model/extensions.scala
@@ -1,0 +1,46 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package observe.model
+
+import cats.syntax.option.*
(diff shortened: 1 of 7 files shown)


rpiaggio:
Now that the `observationsByWorkflowState` query is working in the ODB, we use that to get the candidate observations, with the same filters used in navigate.

Said query is only available for the STAFF role, so now the role is enforced in observe, even in dev.

Finally, there is a refactor regarding the model extensions: now the extensions are all in their own file, separate from `package`.

toddburnside (commented):
LGTM

rpiaggio (approved):
LGTM :D

PR #1630, opened 2025-02-17
Fix glitches during sequence execution

rpiaggio:
The first commit is just migrating the whole web app to monadic hooks and splitting code in the sequence component to make it more readable/maintainable.

The actual fixes are the last 3 commits, which are pretty small each.


toddburnside (approved):
LGTM

PR #1633, opened 2025-02-21
Exposure controls fixes

rpiaggio:
Step pause/resume/stop/abort were disabled if a sequence stop was requested after the current step. This was wrong and is now fixed: the user could have requested a stop after the current step and still want to pause/resume/stop/abort the current exposure.

I've also renamed things: `ObserveStage.Acquiring` indicated that we were acquiring an image/spectroscopy, but it led to confusion regarding target acquisition. It was thus renamed to `ObserveStage.Exposure`.

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
@@ -167,7 +167,7 @@ object MainApp extends ServerEventHandler:
   //   (6) Initialize ODB client WebSocket connection
   //   (7) Query ready observations on ODB
   private val component =
-    ScalaFnComponent[Unit]: _ =>
+    ScalaFnComponent[Unit] { _ =>
       for
         toastRef         <- useToastRef
         syncStatus       <- useStateView(none[SyncStatus]) // UI is synced with server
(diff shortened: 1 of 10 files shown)


modules/web/client/src/main/scala/observe/ui/components/MainApp.scala
```diff
@@ -167,7 +167,7 @@ object MainApp extends ServerEventHandler:
   //   (6) Initialize ODB client WebSocket connection
   //   (7) Query ready observations on ODB
   private val component =
-    ScalaFnComponent[Unit]: _ =>
+    ScalaFnComponent[Unit] { _ =>
```
rpiaggio, line 170: Braces now required to keep scalafix happy

modules/web/server/src/main/resources/app.conf
```diff
@@ -35,7 +35,7 @@ web-server {
     external-base-url = "local.lucuma.xyz:8081"
 
     tls {
-        key-store = "modules/web/server/cacerts.jks.dev"
+        key-store = "cacerts.jks.dev"
```
rpiaggio, line 38: Unrelated, server seems to already resolve the path.

jluhrs: Not on my workspace

rpiaggio:
> Weren't all the `isStopping` also related to the stop button changing and the functionality to cancel the stop ?

isStopping is true when the sequence pause button has been clicked, indicating the sequence to stop after the current step. Still, all the exposure controls for the current step should still be valid.

jluhrs (approved):
LGTM

PR #1638, opened 2025-02-26
SC-4865 Fixed extraction of shutter configuration for GMOS.

jluhrs:
There was a bug in the code from when it was converted to the new OCS model.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/gmos/Gmos.scala
+++ b/modules/server_new/src/main/scala/observe/server/gmos/Gmos.scala
@@ -187,7 +187,7 @@ object Gmos {
   def shutterStateObserveType(obsType: StepType): ShutterState = obsType match {
     case StepType.DarkOrBias(_) | StepType.ExclusiveDarkOrBias(_) | StepType.DarkOrBiasNS(_) =>
       ShutterState.CloseShutter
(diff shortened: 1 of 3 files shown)


modules/server_new/src/main/scala/observe/server/gmos/Gmos.scala
```diff
@@ -187,7 +187,7 @@ object Gmos {
   def shutterStateObserveType(obsType: StepType): ShutterState = obsType match {
     case StepType.DarkOrBias(_) | StepType.ExclusiveDarkOrBias(_) | StepType.DarkOrBiasNS(_) =>
       ShutterState.CloseShutter
-    case _                                                                                   => ShutterState.UnsetShutter
+    case _                                                                                   => ShutterState.OpenShutter
```
cquiroz, line 190: What does it mean `UnsetShutter`? Is it ever used?

cquiroz (approved):
LGTM

modules/server_new/src/main/scala/observe/server/gmos/Gmos.scala
jluhrs: I don't think so. In the old model the value was a string, and the line `case _ => ShutterState.UnsetShutter` was there just to complete the pattern

PR #1639, opened 2025-02-27
Modified code to use enums with Enumerated.

jluhrs:
Modified code to use `enum` and automatically derive `Enumerated`.


cquiroz (approved):
LGTM

PR #1650, opened 2025-03-02
Filter observations that have a reference

rpiaggio:


cquiroz:
Do eng observations have a reference?

rpiaggio:
> Do eng observations have a reference?

I don't know. Shane suggested we use this as filter. @swalker2m ?

swalker2m:
> > Do eng observations have a reference?
> 
> I don't know. Shane suggested we use this as filter. @swalker2m ?

Yes, if the program reference is set (see `Mutation` -> `setProgramReference`).  For example, the format looks like: `G-2025A-ENG-GMOSN-01`.

swalker2m (approved):
:+1:

PR #1652, opened 2025-03-02
Remove refresh button

rpiaggio:
The "Refresh sequence" button was problematic as it was: since the sequence was reloaded in the server, it would produce a new visit and therefore reset a running sequence to acquisition, even if it was executing science already.

If we want to support refresh, we need an endpoint in the server that reloads the current atom without creating a new visit or resetting breakpoints.

Do we want to support this?
- If YES: I'll create a story and leave the button commented out, so that we can reuse it when we have server support.
- If NO: I'll remove the button and all the logic and styling.

### DIFF
--- a/modules/web/client/src/main/scala/observe/ui/components/sequence/SeqControlButtons.scala
+++ b/modules/web/client/src/main/scala/observe/ui/components/sequence/SeqControlButtons.scala
@@ -91,17 +91,17 @@ object SeqControlButtons
             tooltipOptions = tooltipOptions,
             onClick = sequenceApi.cancelPause(props.obsId).runAsync,
             disabled = props.isCancelPauseInFlight || props.isWaitingNextAtom
-          ).when(selectedObsIsLoaded && props.isRunning && props.isUserStopRequested),
-          Button(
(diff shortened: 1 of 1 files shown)


rpiaggio:
After a bit of thought, I would argue that we do want the reload functionality in the server API:

- If I understand correctly, reload on the server is already implemented and is invoked after every step. We just have to expose this on the API.
- The current auto reload in the server is not enough to keep the current atom updated: the sequence could be paused and QA of a previous step is set to failed. Currently, the server will not know about this and will not reload.
- If there's a reload endpoint in the API we can invoke it from the UI when QA is set.
- Alternatively, the server can subscribe to dataset changes and reload when there's one. However, in case of the subscription failing and a missed event, the server won't reload. Therefore, we probably want the operator to do manual reloads to get unstuck.

cquiroz:
Could it be that we enable the refresh button when we know there is something to refresh?
Not sure if the ODB can support that

rpiaggio:
> Could it be that we enable the refresh button when we know there is something to refresh? Not sure if the ODB can support that

How would that work?

cquiroz (approved):
Let's just merge this

PR #1655, opened 2025-03-02
Refresh visits and future sequence on dataset edit

rpiaggio:
This will only refresh in the UI the visits and atoms after the current one.

In order to refresh the current atom on a dataset edit, which is desirable since it can create new steps, we need to add this functionality to the server.


cquiroz (approved):
LGTM

PR #1673, opened 2025-03-10
Do not automatically clear last observation and fix glitch when completing sequence

rpiaggio:
This PR implements a more robust stitching algorithm which avoids duplicated steps when wrapping up atoms/sequence.

Also, it shows completed observations in the obs list, which avoids the recently completed sequence from disappearing. We don't want to actually show completed observations in the list, but this is a temporary workaround while I find another solution.

### DIFF
--- a/deploy/src/main/resources/conf/app.conf
+++ b/deploy/src/main/resources/conf/app.conf
@@ -5,7 +5,7 @@
 #
 
 # If mode is DEVELOPMENT or STAGING, UI provides some extra tools
-environment = STAGING
+environment = DEVELOPMENT
 site = ${SITE}
 explore-base-url = "https://explore-dev.lucuma.xyz"
 
--- a/modules/web/client/src/clue/scala/observe/queries/ObsQueriesGQL.scala
+++ b/modules/web/client/src/clue/scala/observe/queries/ObsQueriesGQL.scala
@@ -15,7 +15,7 @@ object ObsQueriesGQL {
     val document = s"""
       query($$site: Site!, $$date: Date!) {
         observationsByWorkflowState(
-          states: [READY, ONGOING],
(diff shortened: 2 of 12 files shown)


rpiaggio:
Fixes https://app.shortcut.com/lucuma/story/4918/brief-glitch-when-completing-sequence

rpiaggio:
I added a new commit that prevents the completed observations from being shown on the table. Thus fixes [sc-3967].

deploy/src/main/resources/conf/app.conf
```diff
@@ -5,7 +5,7 @@
 #
 
 # If mode is DEVELOPMENT or STAGING, UI provides some extra tools
-environment = STAGING
+environment = DEVELOPMENT
```
cquiroz, line 8: why this change?

modules/web/client/src/clue/scala/observe/queries/ObsQueriesGQL.scala
```diff
@@ -15,7 +15,7 @@ object ObsQueriesGQL {
     val document = s"""
       query($$site: Site!, $$date: Date!) {
         observationsByWorkflowState(
-          states: [READY, ONGOING],
+          states: [READY, ONGOING, COMPLETED],
```
cquiroz, line 18: Would this return a lot of completed observations?

rpiaggio: Only the ones that are active and are associated to the current site.

cquiroz (approved):
LGTM

deploy/src/main/resources/conf/app.conf
rpiaggio: The development deployed instances were incorrectly showing STAGING.

PR #1679, opened 2025-03-11
Setting QA for newly created step records used to fail

rpiaggio:
It was using a stale callback, it was now moved to table meta.

Also: improvements in observation table, which used to show some default data instead of the correct one (status, loaded).


cquiroz (approved):
LGTM

PR #1718, opened 2025-03-24
Library updates and keyword change

cquiroz:


modules/server_new/src/test/scala/observe/server/ObserveEngineSuite.scala
```diff
@@ -362,7 +362,7 @@ class ObserveEngineSuite extends TestCommon {
     val resources: Set[Resource | Instrument] = Set(Instrument.GmosNorth, TCS)
 
     val obsTypes: NonEmptyList[(ObserveClass, StepConfig)] = NonEmptyList(
-      (ObserveClass.ProgramCal, StepConfig.Dark),
+      (ObserveClass.NightCal, StepConfig.Dark),
```
swalker2m, line 365: Are darks taken at night?  If it was `ProgramCal` before it should be `NightCal` now so :+1:

cquiroz: I think darks can be taken at any time

cquiroz: The test in particular doesn't care too much about the class, it doesn't check the content of keywords

modules/server_new/src/main/scala/observe/server/keywords/ObsKeywordReader.scala
```diff
@@ -95,12 +95,10 @@ object ObsKeywordReader {
...
       override def obsClass: F[String] = (step.observeClass match {
-        case ObserveClass.Science        => "science"
-        case ObserveClass.ProgramCal     => "progCal"
-        case ObserveClass.PartnerCal     => "partnerCal"
-        case ObserveClass.Acquisition    => "acq"
-        case ObserveClass.AcquisitionCal => "acqCal"
-        case ObserveClass.DayCal         => "dayCal"
+        case ObserveClass.Science     => "science"
+        case ObserveClass.NightCal    => "nightCal"
+        case ObserveClass.Acquisition => "acqCal"
```
swalker2m, line 100: As you mentioned, this should be `acq`.  Otherwise it looks right to me.

toddburnside (approved):
LGTM

PR #1722, opened 2025-03-26
UI rework

rpiaggio:
- Obs list is now a popup
- It has configuration and constraints columns
- Irrelevant columns were removed, as well as unused code inherited from seqexec
- Global filter
- Better catching of obs loading errors

This should probably have been more than one PR, sorry for that.

Kapture 2025-03-26 at 13 08 49


cquiroz (approved):
LGTM

PR #1726, opened 2025-03-26
Reset acquisition sequence on load

rpiaggio:
Fixes https://app.shortcut.com/lucuma/story/4225/include-reset-flag-in-first-query-of-nextatom

Also, reestructures the ObserveEngine test so as to avoid code repetition

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
+++ b/modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
@@ -668,6 +668,7 @@ object ObserveEngine {
       val author = s", by '${user.displayName}' on client ${clientId.value}."
       executeEngine.inject(
         systems.odb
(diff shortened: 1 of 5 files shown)


modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -660,28 +660,28 @@ object ObserveEngine {
...
+      obsId:    Observation.Id,
       observer: Observer,
       user:     User,
       clientId: ClientId
     ): F[Unit] =
       val author = s", by '${user.displayName}' on client ${clientId.value}."
       executeEngine.inject(
         systems.odb
-          .read(sid)
+          .read(obsId, reset = true)
```
cquiroz, line 671: I guess this is the key change. maybe a comment on why it is true would be good here

cquiroz (approved):
LGTM

PR #1730, opened 2025-03-27
Don't force step to running if waiting prompt, renames

rpiaggio:
When the sequence is waiting user prompt, it's state is `Running` but no step is running. The engine used to force the first step to running if the sequence is `Running`, but we don't want this if waiting user prompt.

Also: clarifying renames.

### DIFF
--- a/modules/engine/src/main/scala/observe/engine/EngineStep.scala
+++ b/modules/engine/src/main/scala/observe/engine/EngineStep.scala
@@ -20,7 +20,29 @@ case class EngineStep[F[_]](
   id:         Step.Id,
   breakpoint: Breakpoint,
   executions: List[ParallelActions[F]]
-)
+):
+  /**
+   * Calculate the `Step` `Status` based on the underlying `Action`s.
+   */
+  lazy val status: StepState =
+    // Find an error in the Step
+    executions
+      .flatMap(_.toList)
+      .find(Action.errored)
+      .flatMap: x =>
+        x.state.runState match
+          case ActionState.Failed(Result.Error(msg)) => msg.some
+          case _                                     => None
+          // Return error or continue with the rest of the checks
+      .map[StepState](StepState.Failed.apply)
+      .getOrElse:
+        // All actions in this Step were completed successfully, or the Step is empty.
+        if (executions.flatMap(_.toList).exists(Action.aborted)) StepState.Aborted
+        else if (executions.flatMap(_.toList).forall(Action.completed)) StepState.Completed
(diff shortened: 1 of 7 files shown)


modules/engine/src/main/scala/observe/engine/EngineStep.scala
```diff
@@ -20,47 +20,39 @@ case class EngineStep[F[_]](
...
-
-  def breakpointL[F[_]]: Lens[EngineStep[F], Boolean] =
-    Focus[EngineStep[F]](_.breakpoint).andThen(isoBool)
-
+):
   /**
    * Calculate the `Step` `Status` based on the underlying `Action`s.
    */
-  private def status_[F[_]](step: EngineStep[F]): StepState =
+  lazy val status: StepState =
```
rpiaggio, line 27: This change is unrelated, it's just simplification + memoization.

modules/model/shared/src/main/scala/observe/model/SequenceState.scala
```diff
@@ -52,11 +52,11 @@ enum SequenceState(val name: String) derives Eq, Encoder, Decoder:
       case SequenceState.Running(_, _, _, _) => true
       case _                                 => false
 
-  def isWaitingNextAtom: Boolean =
+  def isWaitingPrompt: Boolean =
```
cquiroz, line 55: Good change, a lot clearer

modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -1706,8 +1706,10 @@ object ObserveEngine {
         // The sequence could be empty
         case Nil => Nil
         // Find first Pending ObserveStep when no ObserveStep is Running and mark it as Running
+        // When waiting user prompt, the sequence is Running, but no steps are.
         case steps
-            if Sequence.State.isRunning(st) && steps.forall(_.status =!= StepState.Running) =>
+            if (Sequence.State.isRunning(st) && !Sequence.State.isWaitingUserPrompt(st)) && steps
```
cquiroz, line 1711: I think this is the key-change. so when waiting no steps are running and it should be equivalent to the previous contidiotn

cquiroz (approved):
LGTM

modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
rpiaggio: Correct

PR #1731, opened 2025-03-28
Avoid acquisition popup after 1st step

rpiaggio:
It turns out we do need a `waitingNextAtom` parameter, but after the latest changes we can't derive whether to show the prompt or not from it.

In this PR, I'm decoupling both concepts and introducing explicit parameters for each.

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/SequenceState.scala
+++ b/modules/model/shared/src/main/scala/observe/model/SequenceState.scala
@@ -90,8 +90,11 @@ object SequenceState:
     val internalStop: Lens[SequenceState.Running, Boolean] =
       Focus[SequenceState.Running](_.internalStop)
 
-    // val isWaitingUserPrompt: Lens[SequenceState.Running, Boolean] =
-    //   Focus[SequenceState.Running](_.isWaitingUserPrompt)
+    val waitingUserPrompt: Lens[SequenceState.Running, Boolean] =
+      Focus[SequenceState.Running](_.waitingUserPrompt)
 
-    // val isStarting: Lens[SequenceState.Running, Boolean] =
-    //   Focus[SequenceState.Running](_.isStarting)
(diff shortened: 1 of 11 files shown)


toddburnside (approved):
LGTM

modules/model/shared/src/main/scala/observe/model/SequenceState.scala
```diff
@@ -17,24 +17,25 @@ import monocle.macros.GenPrism
 enum SequenceState(val name: String) derives Eq, Encoder, Decoder:
   case Idle                extends SequenceState("Idle")
   case Running(
-    userStop:            Boolean,
-    internalStop:        Boolean,
-    isWaitingUserPrompt: Boolean,
-    isStarting:          Boolean
+    userStop:          Boolean,
```
cquiroz, line 20: It maybe good to document these fields. Especially as they are all booleans

modules/model/shared/src/main/scala/observe/model/SequenceState.scala
```diff
@@ -17,24 +17,25 @@ import monocle.macros.GenPrism
...
+    internalStop:      Boolean,
+    waitingUserPrompt: Boolean,
+    waitingNextAtom:   Boolean,
+    starting:          Boolean
```
cquiroz, line 24: Is this true only once?

modules/model/shared/src/main/scala/observe/model/SequenceState.scala
rpiaggio: I'll switch to newtypes in a future PR.

modules/model/shared/src/main/scala/observe/model/SequenceState.scala
rpiaggio: Whenever `start` is invoked in `ObserveEngine` it's set to true.

PR #1736, opened 2025-03-31
Implement obs table suggestions

rpiaggio:
- Remove obs. name column
- Ensure pop up close icon is always shown
- Change target column filter to simple text

As requested in https://app.shortcut.com/lucuma/story/5010/use-pop-up-observation-chooser-in-observe


cquiroz (approved):
LGTM

PR #1742, opened 2025-04-04
Migrate to latest core

rpiaggio:


modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
```diff
@@ -29,11 +29,11 @@ class ObserveCommandRoutes[F[_]: Async: Compression](
   oe:        ObserveEngine[F]
 ) extends Http4sDsl[F] {
 
-  given EntityDecoder[F, WaterVapor]      = jsonOf[F, WaterVapor]
-  given EntityDecoder[F, ImageQuality]    = jsonOf[F, ImageQuality]
-  given EntityDecoder[F, SkyBackground]   = jsonOf[F, SkyBackground]
-  given EntityDecoder[F, CloudExtinction] = jsonOf[F, CloudExtinction]
-  given EntityDecoder[F, List[Step.Id]]   = jsonOf[F, List[Step.Id]]
+  given wvDecoder: EntityDecoder[F, WaterVapor]             = jsonOf[F, WaterVapor]
```
cquiroz, line 32: Why do you need to name the givens?

cquiroz (approved):
LGTM

modules/web/server/src/main/scala/observe/web/server/http4s/ObserveCommandRoutes.scala
rpiaggio: Because the automated names now conflict since they are generated with non-fully-qualified type. So, `Preset` conflicts. I thought everything looked nicer if I named everything and not only one or two.

PR #1766, opened 2025-04-17
Always show toast when loading fails

rpiaggio:
Loading errors are usually shown in the main screen but sometimes that can fail if the UI's query to the ODB is held.

This PR makes sure the toast is always shown, just in case.


cquiroz (approved):
LGTM

PR #1767, opened 2025-04-18
Emit StepCompleted event distinct from Executed

rpiaggio:
It seems that in `seqexec` an `Exectued` event signalled that a step was complete. But that does not seem to be the case in `observe`, where the step execution is broken up into several `Execution`s.

However, it seems that some parts of the engine still relied on the old logic and things that should be checked when a step ends were checked on `Executed`, resulting in early triggering. In particular, this was the case with processing a requested pause after the current step.

This is now fixed by checking the requested pause only after the last action. Some duplicated code was removed in the process.

Furthermore, an new `StepCompleted` system event is introduced, which is propagated to the clients. This is necessary to play an audible signal on step completion, which is also implemented in this PR.


rpiaggio:
The changes in the web module are somewhat independent of the ones in the server. I recommend reviewing the server first and the ui afterwards.

modules/engine/src/main/scala/observe/engine/Engine.scala
```diff
@@ -140,67 +140,47 @@ class Engine[F[_]: MonadThrow: Logger, S, U] private (
       _.map { seq =>
         seq.status match {
           case SequenceState.Running(userStop, internalStop, _, _, _) =>
-            if (userStop || internalStop) {
```
cquiroz, line 143: it seems we don't care about `userStop` anymore?

rpiaggio: `userStop` seems to be true when the user requests a pause after the current step completes.

modules/engine/src/main/scala/observe/engine/Engine.scala
```diff
@@ -275,7 +255,10 @@ class Engine[F[_]: MonadThrow: Logger, S, U] private (
             // The sequence is marked as completed here
             putS(id)(seq) *> send(finished(id))
           case seq @ Sequence.State.Zipper(z, _, _) =>
-            val stepId                        = z.focus.toStep.id
+            val stepId = z.focus.toStep.id
+
+            println(s"CURRENT ACTIONS: ${seq.current.actions}")
```
cquiroz, line 260: println

cquiroz:
I think this should be reviewed by @jluhrs

rpiaggio:
I'll merge since it's approved via #1769 . @jluhrs if you can still take a look at this, I'll appreciate it.

PR #1775, opened 2025-04-21
Use test sso server

cquiroz:


rpiaggio:
Just in case: This won't change the deployment.

The only app.conf files that are deployed in the docker image are:
- `modules/web/server/src/main/resources/app.conf` (this is the base one and the only one used when deploying locally).
- `deploy/src/main/resources/app.conf` (this is the one used for all deployments, including Heroku).

All the other ones are specific to each test and production server and should be copied manually. They are in this repo so they are not lost in case of server failure.

I understand all of this is confusing, and there are historical reasons (at some point we created different docker images for each server).

I will do a PR reorganizing things and adding comments so that it's less confusing.

Also, I'll change the deploy scripts in each server so that they pull their `app.conf` from this repo and add comments to discourage manually editing in the server.

PR #1795, opened 2025-04-29
Revise deployment method

rpiaggio:
Implements a new deployment method where a different image is released for each server.

There's a base image, which has all the logic, and then 6 additional images that just provide configuration for every combination of [GN | GS] and [Heroku | Staging | Production].

This way, all server configurations are kept concentrated in this repo and under source control. Each server only needs to configure locally its secrets.

The set of scripts we had in each server is copied here and updated.


dngomez (approved):
LGTM!

PR #1800, opened 2025-04-30
Fix gs dhs server, bump libs

rpiaggio:


project/plugins.sbt
```diff
@@ -13,3 +13,6 @@ addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.11.1")
 
 // Generate code for GraphQL queries
 addSbtPlugin("edu.gemini" % "sbt-clue" % "0.43.1")
+
+// TODO Remove this when we can upgrade in sbt-lucuma
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.19.0")
```
rpiaggio, line 18: We need to upgrade or the project won't link but I'm having issues with releasing `sbt-lucuma`, which is where this setting should be. As soon as we can fix that, I'll remove from here.

PR #1805, opened 2025-05-01
Re-complete sequence if ODB removes remaining steps

rpiaggio:
The ODB may re-create steps if their QA is set to Usable/Failed. Handling of this case was addressed in a previous PR.

This PR addresses the case where such a step if now set to Pass or QA unset, where the ODB will remove the step from the sequence, and no remaining steps are now left, effectively completing the sequence.


cquiroz (approved):
LGTM

PR #1807, opened 2025-05-02
Consolidate all site configs in one image

rpiaggio:
Changes the deployment process so that we only release a single image containing all the configurations for all deploy sites. We then select which one to use at each site with an env variable. This results in a simpler process.

Thanks to @hugo-vrijswijk for the idea.


cquiroz:
Is the documentation up to date?

rpiaggio:
> Is the documentation up to date?

Good catch. It wasn't, now it is.

cquiroz (approved):
LGTM

PR #1819, opened 2025-05-16
Logback configuration for deployments

rpiaggio:


modules/web/server/src/main/resources/logback.xml
```diff
@@ -1,81 +1,31 @@
...
- File based logs are rotated daily up to 90 days
+ File based logs are rotated daily up to 90 days.
 
- It supports auto reloading scanning every 60s
+ It supports auto reloading scanning every 60s.
 -->
 <configuration scan="true" scanPeriod="60 seconds">
-  <logger name="org.http4s" level="INFO"/>
-
-  <if condition='property("HOSTNAME").contains("observe")'>
```
cquiroz, line 14: Since you're removing this (good call) we maybe able to remove janino from the dependencies

rpiaggio: Good idea done 👍

cquiroz (approved):
LGTM

PR #1820, opened 2025-05-16
Allow deactivating sound, breakpoint message same as sequence paused

rpiaggio:
image


modules/web/client-model/src/main/scala/observe/ui/model/RootModel.scala
```diff
@@ -92,7 +93,8 @@ object RootModelData:
       observer = none,
       operator = none,
       userSelectionMessage = none,
-      globalLog = FixedLengthBuffer.unsafe(MaxGlobalLogEntries)
+      globalLog = FixedLengthBuffer.unsafe(MaxGlobalLogEntries),
+      isAudioActivated = IsAudioActivated.True
```
cquiroz, line 97: this should be false always :D

cquiroz (approved):
LGTM

PR #1825, opened 2025-05-27
Process Flamingos2 sequences

rpiaggio:
[sc-5704]

Kapture 2025-05-26 at 21 26 50

### DIFF
--- a/modules/model/shared/src/main/scala/observe/model/InstrumentDynamicConfig.scala
+++ b/modules/model/shared/src/main/scala/observe/model/InstrumentDynamicConfig.scala
@@ -5,37 +5,65 @@ package observe.model
 
 import cats.Eq
 import cats.syntax.either.*
+import cats.syntax.eq.*
 import io.circe.Decoder
 import io.circe.DecodingFailure
 import io.circe.Encoder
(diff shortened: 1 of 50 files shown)


modules/model/shared/src/main/scala/observe/model/InstrumentDynamicConfig.scala
```diff
@@ -5,59 +5,95 @@ package observe.model
...
+    type D = gmos.DynamicConfig.GmosSouth
+  }
+
+  case class Flamingos2(config: Flamingos2DynamicConfig)
+      extends InstrumentDynamicConfig(Instrument.Flamingos2) {
+    type D = Flamingos2DynamicConfig
+  }
+
+  object GmosNorth:
+    given Eq[GmosNorth] = Eq.by(x => (x.instrument, x.config))
```
cquiroz, line 46: do we need to compare instrument as it is essentially fixed ?

rpiaggio: Maybe it's not necessary, but I'll leave it for extra caution.

PR #1827, opened 2025-05-27
Readability refactor

rpiaggio:
As a newcomer to observe server, I found parts of the code a bit hard to follow and modify.

This refactor hopes to alleviate the issue for future newcomers and maintainers.

Namely:
- Some type aliases were named `FooType`, which always tripped me into believing that they represented some kind of type, when they represent a `Foo`. `Type` was thus removed from the aliases, which I believe makes code easier to follow.
- Switched to always use explicit `Handle` constructors, rather than relying on `(S => S).toHandle` extension methods over functions.
- `Handle` itself was refactored a bit, renaming the type parameters to the first letter of what they represent, adding convenience methods, and renaming others so that they are more descriptive.
- `Engine`'s method aliases for type-instantiated `Handle` were moved to an encapsulated `EngineHandle` object.


rpiaggio:
Ok I went ahead and decided go all in:

I unified the `engine` and `server` projects, dropping the `S` and `U` type-parameters in the process and hard-coding them.

I understand these came from the original desire to have a generic `Engine` but that has not been the case for a long time and achieving that goal would probably require almost a full rewrite. In the meantime, this refactor significantly simplifies code, removing a lot of type params from everywhere.

modules/server_new/src/main/scala/observe/server/ObserveEngine.scala
```diff
@@ -558,50 +540,71 @@ object ObserveEngine {
...
-              // Read the next atom from the odb and replaces the current atom
-              val atomGen: Option[AtomGen[F]] = translator.nextAtom(odbObs, atomType)._2
-              Event
-                .modifyState[F, EngineState[F], SeqEvent]({ (oldState: EngineState[F]) =>
+  ): EngineHandle[F, Unit] =
+    EngineHandle.fromSingleEventF {
+      Logger[F].debug(s"Reloading atom for observation [$obsId]") >>
+        odb
+          .read(obsId)
+          // TODO Simulate error.
```
cquiroz, line 559: ??? Why do we need to simulate an error here? Maybe this belongs to an instrument simulator?

modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
```diff
@@ -66,13 +65,15 @@ import SeqEvent.*
...
 private class ObserveEngineImpl[F[_]: Async: Logger](
-  executeEngine:                    Engine[F, EngineState[F], SeqEvent],
+  executeEngine:                    Engine[F],
   override val systems:             Systems[F],
   @annotation.unused settings:      ObserveEngineConfiguration,
   translator:                       SeqTranslate[F],
   @annotation.unused conditionsRef: Ref[F, Conditions]
 )(using Monoid[F[Unit]])
     extends ObserveEngine[F] {
+  // import executeEngine.EngineHandle
```
cquiroz, line 75: Commented Code

modules/server_new/src/main/scala/observe/server/engine/Engine.scala
```diff
@@ -0,0 +1,596 @@
...
+                      ) *>
+                        send(Event.executing(id))
+                }
+              }
+            case _                                                                  => EngineHandle.unit
+          }
+        }.getOrElse(EngineHandle.unit)
+      )
+
+  /**
```
cquiroz, line 232: Double set of comments?

modules/server_new/src/main/scala/observe/server/engine/Engine.scala
```diff
@@ -0,0 +1,596 @@
...
+            val w: List[EngineHandle[F, Unit]] =
+              seq.current.actions.indices
+                .map(i => EngineHandle.modifySequenceState[F](obsId)(_.start(i)))
+                .toList
+            w.sequence *> Handle.fromEventStream(v)
+        }
+        .getOrElse(EngineHandle.unit)
+    )
+
+  // private def getState(f: EngineState[F] => Stream[F, Event[F]]): EngineHandle[F, Unit] =
```
cquiroz, line 285: Commented code

modules/server_new/src/main/scala/observe/server/engine/Engine.scala
```diff
@@ -0,0 +1,596 @@
...
+    onSystemEvent: PartialFunction[SystemEvent, EngineHandle[F, Unit]]
+  )(ev: Event[F])(using Concurrent[F]): EngineHandle[F, EventResult] =
+    ev match
+      case Event.EventUser(ue)   => handleUserEvent(ue)
+      case Event.EventSystem(se) =>
+        handleSystemEvent(se).flatMap: (r: EventResult) =>
+          onSystemEvent.applyOrElse(se, (_: SystemEvent) => EngineHandle.unit).as(r)
+
+  /** Traverse a process with a stateful computation. */
+  // input, stream of events
```
cquiroz, line 520: Are these comments for the parameters?

modules/server_new/src/main/scala/observe/server/engine/Event.scala
```diff
@@ -0,0 +1,126 @@
...
+    EventUser[F](LogError(msg, ts))
+  def logErrorMsgF[F[_]: Sync](msg: String): F[Event[F]]                        =
+    Sync[F].delay(Instant.now).map(t => EventUser[F](LogError(msg, t)))
+
+  def pure[F[_]: Sync](v: SeqEvent): Event[F] = EventUser[F](Pure(v))
+
+  def failed[F[_]](id: Observation.Id, i: Int, e: Result.Error): Event[F]  =
+    EventSystem[F](Failed(id, i, e))
+  def completed[F[_], R <: Result.RetVal](
+    id:     Observation.Id,
```
cquiroz, line 77: Could we either make all parameters have a meaningful name or none?

modules/server_new/src/main/scala/observe/server/engine/EventResult.scala
```diff
@@ -0,0 +1,20 @@
...
+sealed trait EventResult extends Product with Serializable
+
+object EventResult:
+  enum Outcome:
+    case Ok, Failure
+
+  case class UserCommandResponse[F[_]](
+    userEvent: UserEvent[F],
+    outcome:   Outcome,
+    ud:        Option[SeqEvent]
```
cquiroz, line 17: ud?

modules/server_new/src/main/scala/observe/server/engine/Handle.scala
```diff
@@ -0,0 +1,112 @@
...
+ * @tparam F[_]:
+ *   Type of the effect
+ * @tparam S:
+ *   Type of the state machine state.
+ * @tparam E:
+ *   Type of the events
+ * @tparam O:
+ *   Type of the output (usually Unit)
+ */
+final case class Handle[F[_], S, E, O](run: StateT[F, S, (O, Stream[F, E])])
```
cquiroz, line 27: I think this is the core change. Could you describe it a little bit more?

modules/server_new/src/main/scala/observe/server/engine/Handle.scala
```diff
@@ -0,0 +1,112 @@
...
+      override def flatMap[O, O1](
+        fa: Handle[F, S, E, O]
+      )(f: O => Handle[F, S, E, O1]): Handle[F, S, E, O1] =
+        Handle[F, S, E, O1]:
+          fa.run.flatMap: (a, p1) =>
+            f(a).run.map: (b, p2) =>
+              (b, p1 ++ p2)
+
+      // Kudos to @tpolecat
+      def tailRecM[O, O1](a: O)(f: O => Handle[F, S, E, Either[O, O1]]): Handle[F, S, E, O1] = {
```
cquiroz, line 66: Nightmares from the past...

modules/server_new/src/main/scala/observe/server/engine/package.scala
```diff
@@ -0,0 +1,120 @@
...
+extension [F[_]](v: List[Action[F]]) {
+  def prepend(ac: List[ParallelActions[F]]): List[ParallelActions[F]] =
+    NonEmptyList.fromList(v).foldRight(ac)(_ :: _)
+}
+
+enum OnAtomReloadAction:
+  case NoAction, StartNewAtom
+
+// Some types defined to avoid repeating long type definitions everywhere
+// type EngineEvent[F[_]]     = Event[F]
```
cquiroz, line 44: Commented code

cquiroz (commented):
This looks like a very good improvement. Thank you for taking this task. 
Since you have tested it, it seems fine to me. Okay, perhaps we need @jluhrs  to take a look.

modules/server_new/src/main/scala/observe/server/engine/EventResult.scala
rpiaggio: I'd love to know what `ud` means so that we can rename it.

modules/server_new/src/main/scala/observe/server/engine/Engine.scala
rpiaggio: I'm assuming. I didn't change that.

modules/server_new/src/main/scala/observe/server/engine/Event.scala
rpiaggio: The goal is to have meaningful names everywhere, but that may take time. I've now renamed these.

modules/server_new/src/main/scala/observe/server/engine/Handle.scala
rpiaggio: In the class signature the changes were:
- Renaming type parameters so that `S`: State, `E`: Event, `O`: Output.
- The event stream is no longer `Option`al. We use `Stream.empty` if there are no events. This generally simplifies code.

There were some changes to constructors:
- `(S => (S, O)).toHandle` is now `Handle.modifyState(S => (S, O))`.
- `(S => F[(S, O)]).toHandleF` is now `Handle.modifyStateF(S => F[(S, O)])`.
- `(StateT).toHandle` is now `Handle.fromStateT(StateT)`.
- `modifyState(S => S)` is now `modifyState_(S => S)` (using `_` for an output of `Unit`, as is conventional).
- `get`, `inspect`, `modify` are now `getState`, `inspectState`, `modifyState`.
- `fromEventStream` now has variants accepting `(S => Stream[F, E])`, just `Stream[F, E]` and `fromSingleEventF(F[E])` and `fromSingleEvent(E)`.

Some of these were in other classes or in extensions in different parts. Now it's all concentrated here.

modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
```diff
@@ -422,55 +413,54 @@ private class ObserveEngineImpl[F[_]: Async: Logger](
...
 //      i:        Instrument,
 //      sid:      Observation.Id,
 //      observer: Observer,
 //      user:     User,
 //      clientId: ClientId
-//    ): EventType[F] = {
+//    ): Event[F] = {
 //      val lens = EngineState
 //        .sequences[F]
 //        .andThen(mapIndex[Observation.Id, SequenceData[F]].index(sid))
```
jluhrs, line 466: Why keep these commented lines?

modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
```diff
@@ -1306,7 +1274,7 @@ private class ObserveEngineImpl[F[_]: Async: Logger](
 //                    if q.status(st) =!= BatchExecState.Completed =>
 //                  runNextsInQueue(qid, observer, user, clid, freed).flatMap { l =>
 //                    Handle.fromStream(
-//                      Stream.emit[F, EventType[F]](
+//                      Stream.emit[F, Event[F]](
 //                        Event.modifyState[F, EngineState[F], SeqEvent](
 //                          executeEngine.pure(SequencesStart(l))
 //                        )
```
jluhrs, line 1280: And these lines?

modules/server_new/src/main/scala/observe/server/engine/EventResult.scala
jluhrs: User something?

modules/server_new/src/main/scala/resources/Altair.xml
jluhrs, line 1: Why was the `resources` folder moved inside the `scala` folder ?

modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
rpiaggio: I left all the lines that were commented assuming that's code from seqexec that we will use again at some point. Should I remove them?

modules/server_new/src/main/scala/resources/Altair.xml
rpiaggio: Oops, I must have done this inadvertently. Good catch. Reverting.

jluhrs (approved):
LGTM

PR #1831, opened 2025-05-29
Move systems configuration to local file (systems.conf)

rpiaggio:
Also, access ODB through https instead of http.

### DIFF
--- a/deploy/conf/staging-gs/site.conf
+++ b/deploy/conf/staging-gs/site.conf
@@ -44,23 +44,8 @@ observe-engine {
   #   full: connect and command the system
   #   readOnly: connect, but only to read values
   #   simulated: don't connect, simulate internally
-  # You can override in local secrets.conf
   system-control {
-      dhs = full
-      flamingos-2 = full
-      gcal = full
-      ghost = simulated
-      ghost-gds = simulated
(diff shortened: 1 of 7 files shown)


modules/web/server/src/main/resources/base.conf
```diff
@@ -50,6 +50,7 @@ observe-engine {
     system-control {
         dhs = simulated
         flamingos-2 = simulated
+        ig2 = simulated
```
cquiroz, line 53: igirns2 and igrins2Gds

deploy/conf/staging-gs/site.conf
```diff
@@ -44,23 +44,8 @@ observe-engine {
...
-      gnirs = simulated
-      gpi = simulated
-      gpi-gds = simulated
-      gsaoi = simulated
-      gws = full 
-      nifs = simulated
-      niri = simulated
-      tcs = full
-      gems = simulated
+      # Systems should be configured in local systems.conf file.
```
cquiroz, line 48: is the fallback all simulated?

rpiaggio: yes, it's the one in `base.conf`; the same one that's used in dev

PR #1835, opened 2025-06-04
Make Handle an opaque type, make it a MonadThrow, add law test

rpiaggio:
- `Handle` is now an `opaque type` instead of a `case class`, which avoids allocation.
- It can now handle errors in-monad since it implements `MonadThrow`.
- Added law tests.
- Fixed `tailRecM` implementation to make it lawful.

### DIFF
--- a/modules/server_new/src/test/scala/observe/server/engine/HandleSpec.scala
+++ b/modules/server_new/src/test/scala/observe/server/engine/HandleSpec.scala
@@ -1,0 +1,69 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package observe.server.engine
+
+import cats.FlatMap
+import cats.Id
+import cats.Monad
+import cats.data.StateT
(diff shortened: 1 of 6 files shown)


modules/server_new/src/test/scala/observe/server/engine/HandleSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+import observe.engine.Handle
+import org.scalacheck.Arbitrary
+import org.scalacheck.Arbitrary.arbitrary
+import org.scalacheck.Cogen
+
+import scala.util.Try
+
+
+class HandleSpec extends munit.DisciplineSuite {
+  given [F[_]: Monad, S, E, O](using
```
cquiroz, line 26: Probably not needed but should these be on its own arb class?

cquiroz (approved):
LGTM

modules/server_new/src/test/scala/observe/server/engine/HandleSpec.scala
rpiaggio: I thought about it and decided that we can split it if we ever need it.

PR #1837, opened 2025-06-05
Temporary workaround for Gaia timeouts + DHS simulator fix

rpiaggio:
- Use WS client to communicate with ODB. This gets us around Heroku's 30s timeout.
- Increase engine heartbeat timeout. This seemed to be an issue, not sure why.
- Switch DHS simulator to seconds instead of nanos, we were overflowing the Long.


modules/server_new/src/main/scala/observe/server/keywords/DhsClientHttp.scala
```diff
@@ -249,11 +249,11 @@ private class DhsClientSim[F[_]: FlatMap: Logger](site: Site, date: LocalDate, c
 object DhsClientSim {
   def apply[F[_]: Sync: Logger](site: Site): F[DhsClient[F]] =
     Clock[F].monotonic
-      .map(d => Instant.EPOCH.plusNanos(d.toNanos))
+      .map(d => Instant.EPOCH.plusSeconds(d.toSeconds))
```
cquiroz, line 252: 👍

cquiroz (approved):

PR #1838, opened 2025-06-06
Notify UI of errors even if odb record fails

rpiaggio:
When there's an execution error we:
- Notify the ODB.
- Notify the UI.

The UI notification was subject to the ODB notification succeeded. This PR disentangles them, making both in parallel. If the ODB notification fails, the UI will receive a 2nd notification informing of this.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
+++ b/modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
@@ -888,15 +888,20 @@ private class ObserveEngineImpl[F[_]: Async: Logger](
         .as[TargetedClientEvent](BaDum)
     ) ++
       stream(EngineState.default[F])
-        .flatMap: x =>
-          Stream.eval(notifyODB(x).attempt)
-        .flatMap:
-          case Right((ev, qState)) => toClientEvent(ev, qState, systems.odb)
-          case Left(e)             =>
-            Stream.eval:
-              LogMessage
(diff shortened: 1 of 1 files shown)


modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
```diff
@@ -1344,15 +1349,15 @@ private class ObserveEngineImpl[F[_]: Async: Logger](
     i: (EventResult, EngineState[F])
   ): F[(EventResult, EngineState[F])] =
     (i match {
-      case (SystemUpdate(SystemEvent.Failed(id, _, e), _), _) =>
-        Logger[F].error(s"Error executing $id due to $e") <*
+      case (SystemUpdate(SystemEvent.Failed(obsId, _, e), _), _) =>
+        Logger[F].error(s"Error executing $obsId due to $e") <*
           systems.odb
-            .stepAbort(id)
+            .stepAbort(obsId)
```
cquiroz, line 1355: Does this notify the client? I think the client wants the obs reference to be displayed rather than the id

cquiroz (approved):
LGTM

modules/server_new/src/main/scala/observe/server/ObserveEngineImpl.scala
rpiaggio: This is the ODB notification. I agree with you, I'll look into it in another PR.

PR #1840, opened 2025-06-06
Avoid scrolling to top of table between steps, fix DhsClientDisabled

rpiaggio:
There's a brief moment where there's no selected row. This caused the sequence to jump to the first row.

Also: Fix DhsClientDisabled.


modules/web/client/src/main/scala/observe/ui/components/sequence/SequenceTableBuilder.scala
```diff
@@ -56,17 +56,19 @@ private trait SequenceTableBuilder[S: Eq, D <: DynamicConfig: Eq] extends Sequen
...
-          _.scrollToIndex(
-            table
-              .getRowModel()
-              .rows
-              .indexWhere(row => rowIdCandidates.contains_(row.id.value)) - 1,
-            ScrollOptions
-          )
-      )
+    rowIdCandidates.some
+      .filter(_.nonEmpty)
```
cquiroz, line 60: tricksy

cquiroz (approved):
LGTM

PR #1874, opened 2025-06-23
Add basic tracing, skip empty streams

rpiaggio:
- Add tracing for HTTP requests to the server, as well as when the server invokes the ODB.
- Avoids processing empty event streams. A recent refactor had introduced empty streams (instead of optional). This may or may not be related to the race-based issue we are getting where the sequence finishes early. It's worth a shot and it's an optimization anyway.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -265,11 +265,12 @@ lazy val observe_server = project
         PPrint.value,
         Clue.value,
         ClueHttp4s,
+        ClueNatchez,
         CatsParse.value,
         ACM,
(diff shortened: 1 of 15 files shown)


cquiroz (approved):
LGTM, you may need to request permission to connect to honeycomb from the ops machines

rpiaggio:
> LGTM, you may need to request permission to connect to honeycomb from the ops machines

Thank you for the heads up, I'll verify.

PR #1877, opened 2025-06-24
Block sequence state updates while refreshing is in progress

rpiaggio:
This seems to solve the issue where sequences unexpectedly ended early.

The issue seemed to be that when the engine was moving from one step to another, the state could sometimes contain invalid data if it was in the middle or a sequence refresh. This is solved by adding a latch (`Deferred[F, Unit]`) to each sequence, which is used to synchronize.

Bonus features:
- `Handle` is now a `MonadCancel` besides a `MonadError`. This allows us to `bracket`, `guarantee`, etc around a `Handle`.
- NewTypes for the 5 boolean members of `SequenceState.Running`.
- `EngineHandle.debug` method that can be easily `flatMap`ped to other `Handle`s.

I'm leaving a lot of the debug info. I would like to deploy this to the servers for further diagnosis in case the problem persists. I'll remove in a future PR if the issue is solved.

Finally, the `waitingNextAtom` mechanism seems to be rendered obsolete. I don't want to touch anything more for the moment, so I will remove and replace for the new one in a future PR.

### DIFF
--- a/modules/server_new/src/main/scala/observe/server/engine/Engine.scala
+++ b/modules/server_new/src/main/scala/observe/server/engine/Engine.scala
@@ -162,7 +162,7 @@ class Engine[F[_]: MonadCancelThrow: Logger] private (
                           userStop,
                           internalStop,
                           waitingUserPrompt = IsWaitingUserPrompt.Yes,
-                          waitingNextAtom = IsWaitingNextAtom.Yes, // TODO Should this be No?
+                          waitingNextAtom = IsWaitingNextAtom.Yes,
                           starting = IsStarting.No
                         )
                       ) *> send(Event.modifyState(atomLoad(this, obsId)))


modules/server_new/src/main/scala/observe/server/engine/Engine.scala
```diff
@@ -137,96 +138,130 @@ class Engine[F[_]: MonadThrow: Logger] private (
...
+                      EngineHandle.debug(s"**** {next} STOPPING SEQUENCE BECAUSE NEXT IS EMPTY") >>
+                        send(Event.finished(obsId))
+                    // Final State
+                    case Some(qs @ Sequence.State.Final(_, _)) =>
+                      EngineHandle.replaceSequenceState(obsId)(qs) *> switch(obsId)(
+                        SequenceState.Running(
+                          userStop,
+                          internalStop,
+                          waitingUserPrompt = IsWaitingUserPrompt.Yes,
+                          waitingNextAtom = IsWaitingNextAtom.Yes, // TODO Should this be No?
```
jluhrs, line 165: Here the current state of the observation is `seq`, and we are checking its future state if we were to progress the execution, `seq.next` (the executin state will not progress until we update the engine state). The fact the `seq.next` is `Final` here means that `seq` doesn't have anything else to execute. Therefore, we need to check  in the ODB if there is something new to execute next, or to declare the atom completed.

rpiaggio: I see, thank you!

I removed the comment.

Anyway, in the future we should remove `waitingNextAtom` and just use the new latch. Otherwise, we have 2 mechanisms for the same goal.

PR #1880, opened 2025-06-24
Solve early completion

rpiaggio:
This PR backtracks on the changes introduced yesterday.

The issue is definitely with the refresh induced by observation edits, but the use of a latch didn't solve the issue since what's actually happening is that the state became inconsistent.

This PR undoes the latching, and permits an observation edit to update the sequence only if it is not running. This seems to work, but let's test it on the servers.


rpiaggio:
I'm still leaving all the debugging info just in case. Will remove if everything works.

cquiroz (approved):
LGTM

PR #1895, opened 2025-06-30
Emit ReadingOut when stopping in pause

rpiaggio:


cquiroz:
Is this trying to account for the fact that at the end of a stop we start reading out?
Shouldn't we get that status from the instrument?

rpiaggio:
> Is this trying to account for the fact that at the end of a stop we start reading out? Shouldn't we get that status from the instrument?

This code executes when Stop is pressed during a step Pause. I agree that the status should be the same as the instrument, let me see if I can find a way to do that.

jluhrs:
The issue is that the process that monitores the exposition progress (the one that gives the countdown and the stage updates) is only activated when starting the exposure and when resuming a paused exposure.
Would it work if it is activated when stopping a paused exposure?

PR #1936, opened 2025-08-01
Use the latest odb schema

cquiroz:


modules/server_new/src/test/scala/observe/server/engine/TestUtil.scala
```diff
@@ -32,7 +32,7 @@ object TestUtil {
             obsData = OdbObservation(
               id = obsId,
               title = NonEmptyString.unsafeFrom("Test Observation"),
-              workflow = OdbObservation.Workflow(ObservationWorkflowState.Ready),
```
toddburnside, line 35: `ObservationWorkflowState` and `CalculationState` are not the same thing. For better or worse, both have a `state` member and both of them have a potential value of `Ready`. 
I think the confusion here is that the `workflow` in the `Observation` query in `ObservationSummarySubquery` no longer returns an `ObservationWorkflow` directly, but a `CalculatedObservationWorkflow`, which is a `CalculatedValue[ObservationWorkflow]`, which has a `state` that shows the state of the calculation by the ObsCalc service, and a `value` which is the old `ObservationWorkflow`.

PR #1962, opened 2025-08-20
Show F2 and GMOS imaging config in observation selector

rpiaggio:
Also restructured code so that it will give an incomplete match error when we add more instruments.


modules/web/client/src/main/scala/observe/ui/components/queue/ObsListPopup.scala
```diff
@@ -110,62 +110,63 @@ object ObsListPopup
...
-        ).sortable.withFilterMethod(FilterMethod.StringText()),
-        ColDef(
-          ConstraintsColumnId,
-          _.constraintsSummary,
-          header = "Constraints",
-          cell = _.value
-        ).sortable.withFilterMethod(FilterMethod.StringSelect())
-      )
+      ): List[ColumnDef[SessionQueueRow, ?, Nothing, WithFilterMethod, String, ?, Nothing]] =
+        List(
```
rpiaggio, line 114: Unrelated change, just formatting.

PR #1970, opened 2025-08-23
Workaround for max memory in heroku+docker woes

rpiaggio:
When deploying to Heroku via Docker, the max memory is reported wrong. It reports the max host memory instead of container's. Thus, using `-XX:MaxRAMPercentage` will allocate beyond the dyno limit and cause the app to crash if it attempts to go beyond such limit.

Although Java 10+ has provisions to detect when it's running in a container and read the available memory from something called `cgroup`, there are `v1` and `v2` versions of `cgroup`. In my local Docker, the container reports the memory as `v2` and the JVM picks it up correctly.

However, in Heroku, apparently due to an outdated Docker container, it's only reported as `v1`. Although the JVM should fall back to `v1` if `v2` is not found (https://bugs.openjdk.org/browse/JDK-8230305), this just isn't happening in Heroku.

I opened up a ticket with Heroku to inquire if there's a solution for this: https://help.heroku.com/sharing/f0782264-57d3-49e0-a381-0c3913376acd

Although the bad news is that the JVM can't set its heap max size automatically, the good news is that correct size is reported somewhere. Thus, I spun a script with the aid of ChatGPT to manually do the grunt work to configure the max heap size according the dyno's limits. This seems to work well.

Unless Heroku support provides a better way of doing this or someone comes up with a better idea, I suggest we use this method throughout.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,7 +13,7 @@ val build = taskKey[File]("Build module for deployment")
 name := "observe"
 
 ThisBuild / dockerExposedPorts ++= Seq(9090, 9091) // Must match deployed app.conf web-server.port
-ThisBuild / dockerBaseImage := "eclipse-temurin:17-jre"
+ThisBuild / dockerBaseImage := "eclipse-temurin:21-jre"
 
 // TODO REMOVE ONCE THIS WORKS AGAIN
 ThisBuild / tlCiScalafmtCheck := false
@@ -352,7 +352,6 @@ lazy val deployedAppSettings = Seq(
   // Launch options
   Universal / javaOptions ++= Seq(
     // -J params will be added as jvm parameters
-    "-J-XX:MaxRAMPercentage=80.0",
     // Support remote JMX access
     "-J-Dcom.sun.management.jmxremote",
     "-J-Dcom.sun.management.jmxremote.authenticate=false",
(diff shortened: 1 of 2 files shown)


toddburnside (approved):
🤯 LGTM. Thanks for figuring out a solution. 
I assume for the itc we will set the HEROKU_JAVA_HEAP_PERCENT to something smaller to account for the magic jar files?

rpiaggio:
Are we positive that those are not loaded into the heap?

If we need to, one advantage of this method is that we can factor in whatever we want when computing the max heap size: for example we can do: 80% - Constant, where Constant would be the size of the objects from the magic jar files.

toddburnside:
> Are we positive that those are not loaded into the heap?

Almost 100%. 😆 They get loaded into some other memory space whose name alludes me at the moment. For the itc I had to drop it's heap size down low to account for the jar files taking more space. It is unclear to me why itc-production is taking more space than dev or staging. I'd have to look at the jar files again, but IIRC their size approaches 400M, with one in particular taking almost all of the space.

toddburnside:
> I would like to put the script in one single place for all the repos to share. I _think_ it suffices for it to be available somewhere in the web and the build can read it from there. Probably the place to do so is in sbt-lucuma. I'll try this. Otherwise, we probably want to bundle the script with the plugin.

Sounds great.

